Issue #13, opened 2026-05-06
Create mac package


Galvill:
## Summary

Ship a first-class macOS distribution so Mac users can install jitenv without building from source.

## Status (2026-05-09)

- **Native macOS port: SHIPPED** in #25 (v0.5.0). Agent runs on darwin/amd64 + darwin/arm64; peer-cred check uses `LOCAL_PEERCRED`; runtime dir falls back to `os.TempDir()` (`/var/folders/.../T/jitenv-<uid>`); `Setsid` double-fork is portable. Goreleaser cross-compiles tarballs.
- **What's left in this issue:** packaging + signing.

## Remaining work

- [ ] Homebrew tap (formula or cask). Likely host at `Galvill/homebrew-jitenv`.
  - Decide: tap (user-installed) vs. submitting to homebrew-core (review queue + maintainer overhead). Start with own tap.
- [ ] Apple notarization. Today release headers tell users to `xattr -d com.apple.quarantine ./jitenv` or right-click → Open. Real fix is paid Developer account + notarytool integration in goreleaser.
- [ ] Optional: signed `.pkg` installer for users who prefer that over an extracted tarball.

## Already done

- [x] jitenv builds and the agent runs correctly on macOS (Intel + Apple Silicon) — #25.
- [x] CI cross-compiles for darwin/amd64 and darwin/arm64 — #25.
- [x] Release artifacts include macOS tarballs — #25.
- [x] Documented in README that the binary isn't notarized + the unquarantine workaround — #25.

## References

- #25 — the macOS port itself, merged.
- `.goreleaser.yaml` — `goos: [linux, darwin]` already set up.
- `internal/agent/peer_darwin.go` — Darwin-specific peer-cred check.

Galvill:
Agent loop paused: issue #13's actionable scope (Homebrew cask tap + Apple notarization) is already implemented and shipped on main via #25/#220/#226/#233/#236. The only remaining item is the *optional* signed .pkg installer, which requires Apple Developer ID credentials (Developer ID cert, Team ID, notary API key) the automated loop does not possess. Decision needed: close #13 as done, or supply signing credentials to scaffold the optional .pkg installer.

PR #19, opened 2026-05-06
docs(design): keychain integration analysis (defer decision)

### LINKED ISSUE #18: Consider OS keychain integration for the master key
## Summary

Investigate whether jitenv should optionally store the master passphrase / derived key in the OS keychain (Linux Secret Service / GNOME Keyring / KWallet, macOS Keychain, Windows Credential Manager) so the user doesn't have to type the passphrase on every `jitenv unlock`.

This is an **investigation / design** issue — not a commitment to ship. The outcome is a design doc and a go/no-go decision.

## Motivation

- UX: typing the passphrase every time is friction, especially on dev machines that reboot frequently.
- Parity with other tools (`gh auth`, `pass`, `op`, etc.) that use the platform keychain.

## Tensions to think through

- **Threat model.** jitenv's design (`CLAUDE.md`: "keeps secrets out of the parent shell", master key zero-on-defer, agent-only decryption) implies a fairly high bar. A keychain-stored key trades passphrase entry for "any process running as the user can ask the keychain" (modulo `kwallet`/`gnome-keyring` per-app prompts). Document explicitly what attacks this does and does not protect against.
- **Opt-in only.** Default behaviour should not change; users who want the extra security of typing a passphrase every time keep it.
- **Cross-platform abstraction.** Consider `99designs/keyring` (Go), or write a thin interface with platform-specific implementations.
- **Headless / CI / SSH.** Keychain is often unavailable or locked; need a clean fallback to passphrase entry.
- **Lock semantics.** `jitenv lock` should still wipe in-memory key. Should it also clear the keychain entry, or only on explicit "forget"?
- **Key vs. passphrase.** Storing the *derived* key skips Argon2id on unlock (faster) but means a keychain compromise yields the working key directly. Storing the *passphrase* preserves the KDF but is slower. Pick one with a written rationale.
- **WSL.** This repo's primary dev platform appears to be WSL — keychain support there is fiddly. Decide whether WSL is in scope for v1 or explicitly out.

## Deliverable

- [ ] Short design doc (in this issue or `docs/design/keychain.md`) covering: threat model, chosen library, key-vs-passphrase decision, opt-in UX, fallback behaviour, platform support matrix.
- [ ] Decision: build it, defer it, or close as won't-do — with reasoning.

Galvill:
## Summary
- Adds `docs/design/keychain.md` with the threat-model analysis, library / key-vs-passphrase decisions, opt-in UX sketch, fallback behaviour and platform matrix for OS keychain storage of the master key.
- Decision: **defer** the build. Revisit once the macOS port (#13) lands so the feature ships with a single coherent threat-model section across all supported platforms.
- Interim UX recommendation noted in the doc: tune the agent idle timeout / surface it in `jitenv status` before introducing keychain.

Closes #18.

## Test plan
- [x] doc-only change; `make build` not affected.
- [ ] Reviewer sanity-check the threat-model section against `CLAUDE.md` security claims.


Galvill:
Brought up to date with `main`. No conflicts (doc-only branch; main doesn't touch `docs/design/keychain.md`).

Re-validated the doc against current state:
- Issue #18 (keychain) — still open.
- Issue #13 (macOS port) — still open, with #25 in flight; the doc's defer-until-macOS rationale still applies.
- Interim recommendation ("tune agent idle timeout") — already partially shipped: `internal/config` has `AgentConfig.IdleTimeout` with a 30m default, surfaced in the TUI Settings screen. Not yet surfaced in `jitenv status` (still shows only `idle for:` elapsed, not the configured ceiling), so that micro-improvement remains a fair follow-up if we want to fully retire the "surface it in status" line.

Ready for review whenever you want to land the design doc.

Issue #17, opened 2026-05-06
Re-add GitHub secrets/variables source with UI-driven, encrypted credentials


Galvill:
## Summary

Bring back a polished GitHub source (Actions secrets / variables and/or Dependabot secrets, scoped to org / repo / environment). Like AWS, the backend is compiled in but the TUI's "Remote Sources" page is hidden — the gap is a first-class UI with **all** configuration managed in the TUI and persisted encrypted in `config.toml`.

## Requirements

- [ ] Re-enable the "Remote Sources" page in the TUI and add a "GitHub" entry.
- [ ] All connection settings configurable in the UI:
  - Personal access token (classic or fine-grained) — encrypted at rest.
  - Optional GitHub App credentials (app id, installation id, private key) as an alternative to PAT — private key encrypted.
  - GitHub host (default `api.github.com`, support GHES).
  - Scope: organization / repository / environment.
- [ ] Variables vs. secrets — note that GitHub does **not** expose secret *values* via the API; only names are listable. Decide UX:
  - Pull *variables* (values readable) and *secret names* (for mapping completeness).
  - For secret values, document that they must be supplied via a separate path (local bag, AWS, etc.) and mapped by name. Make this clear in the UI to avoid confusion.
- [ ] Sensitive fields stored as `enc:v1:` envelopes — never plaintext in `config.toml`.
- [ ] Validate on save: hit `/user` or `/repos/{owner}/{repo}` to confirm the token / app works.
- [ ] Mapping flow consistent with other sources: bind a key, or expand the whole bag.
- [ ] Existing `[sources.*]` entries keep working.

## References

- Source plugin pattern: `pkg/source`, `internal/sources/<name>`, `internal/sources/builtin/builtin.go`.
- Crypto: `internal/crypto`, `DecryptStringsInPlace`.

## Acceptance criteria

- [ ] Round-trip configuration via TUI works on `github.com` and at least one GHES test target.
- [ ] No plaintext token in `config.toml` after save.
- [ ] Clear UX for the "secret values aren't readable" caveat.
- [ ] Documented in README / docs.

Galvill:
Closing as **won't fix**.

Re-evaluated the use case: GitHub's API exposes Actions secret **metadata** but not values. Without secret values, this source collapses to "fetch plaintext variables from GitHub" — too narrow to justify the maintenance cost (a hidden TUI wizard flow, the `google/go-github` dependency, and ongoing re-validation as the rest of the codebase evolves).

[...]

Issue #16, opened 2026-05-06
Re-add AWS Secrets Manager source with UI-driven, encrypted credentials


Galvill:
## Summary

Bring back a polished AWS Secrets Manager source. The backend is already compiled in (per `CLAUDE.md`: "AWS Secrets Manager and GitHub Variables sources are compiled in and work at runtime, but the TUI's 'Remote Sources' page is currently hidden"); the gap is a first-class UI for configuring it, with **all** AWS connection credentials managed in the TUI and persisted encrypted in `config.toml`.

## Requirements

- [ ] Re-enable the "Remote Sources" page in the TUI and add an "AWS Secrets Manager" entry.
- [ ] All connection credentials configurable in the UI, including (at minimum):
  - Access key ID
  - Secret access key
  - Session token (optional)
  - Region
  - Optional assume-role ARN / external ID / session name
  - Optional endpoint override (LocalStack etc.)
- [ ] Sensitive fields (secret access key, session token, external ID) saved as `enc:v1:` envelopes via `crypto.DecryptStringsInPlace` — never stored in plaintext.
- [ ] **No environment-variable or shared-credentials-file fallback** for credentials configured through the UI — the source uses what the user typed, decrypted in the agent only. (Discuss whether to *also* support an "use ambient AWS env / profile" toggle for users who prefer that, but keep UI-config the primary path.)
- [ ] Validate on save: ping STS `GetCallerIdentity` (or similar lightweight call) to give immediate feedback.
- [ ] Mapping flow: pick a secret by name/ARN, then either bind a single key (JSON path) or expand the whole secret as a bag (`Name == ""` semantics).
- [ ] Existing `[sources.*]` entries in users' `config.toml` keep working — no breaking schema changes.

## References

- Constructor / registration pattern: `internal/sources/<name>` + `internal/sources/builtin/builtin.go`.
- Crypto envelope handling: `internal/crypto`, `Meta.Verify`, `DecryptStringsInPlace`.
- Resolver bag injection: `bagSink` in `internal/agent/resolver.go`.

## Acceptance criteria

- [ ] Round-trip: create → save → re-open TUI → connection works without re-entering credentials.
- [ ] `cat config.toml` shows no plaintext AWS secrets.
- [ ] e2e test (or at least an integration test against LocalStack) covers fetch + mapping.
- [ ] Documented in README / docs.

PR #24, opened 2026-05-06
feat(github): polish source schema and add unit tests

Galvill:
## Summary
This is the github-source half of the Remote Sources work for #17. The TUI re-enablement and the shared "Test connection" button are in #16; once both PRs land, the github source is fully UI-managed end-to-end (PAT auth path).

Changes here are deliberately tight:

- **Schema polish.** The `api_url` field's Help text now spells out the GHES `/api/v3/` suffix and surfaces the "GitHub does not expose Actions secret VALUES via the API" caveat **inline in the form**, so the user sees it before saving rather than discovering it at fetch time. The `token` field gets a clearer label ("Personal access token") and explicit "encrypted at rest" wording.
- **Unit tests** for the github source: required-fields validation in `New`, schema shape (which fields are sensitive), `Fetch` validation paths (`ref.Key` required, `scope=secret` rejected with the right error, `scope=repo` requires `owner/repo`, `scope=env` requires `extra.environment`), and `splitOwnerRepo`.

## What's deferred

**GitHub App authentication** (`app_id`, `installation_id`, `private_key`) is listed as **Optional** in the issue, and adding it pulls in `bradleyfalzon/ghinstallation/v2` + `golang-jwt/jwt/v4` as new deps. To keep this PR tight and the dep surface considered, App auth is filed as a follow-up. PAT auth covers the documented happy path and is what most users want.

## Test plan
- [x] `go test ./internal/sources/github/ -v` — 8 cases, all pass.
- [x] `go test ./...` — full suite green.
- [x] `go vet ./...` clean.
- [ ] Reviewer: after #16 lands and Remote Sources is reachable, smoke-test against `github.com` with a fine-grained PAT scoped to one repo's variables.
- [ ] Reviewer: confirm the secrets-not-readable caveat renders in the source params form (focus the API URL field, confirm the Help line shows the NOTE).

## Notes / dependency
- This PR does not modify the TUI menu or any TUI screens. It produces no UI surface on its own — it relies on #16's Remote Sources re-enablement to be reachable from the menu. Both PRs target `main` and don't share files; either order of merging is fine.


Galvill:
Brought up to date with `main`.

Changes:
- Merged `main` (no code conflicts — the github source files this PR touches are untouched on main).
- Fixed the gofmt regression in `github_test.go` (struct-field alignment in `TestSplitOwnerRepo`) that was failing CI.

Verified:
- `gofmt -l ./...` clean.
- `go vet ./...` clean.
- `go test ./...` all green (github source tests: 8 cases pass).

Still relevant: issue #17 (Re-add GitHub secrets/variables source with UI-driven, encrypted credentials) is still open. The corresponding AWS work (#16) merged via #23 in the meantime, so this PR remains the github-source half of that effort. The TUI re-enablement is still a follow-up.

Galvill:
Closing without merging.

The decision has been made not to invest further in the github Variables source: GitHub's API only exposes Actions secret **metadata**, not the actual secret values, so the source can fetch variables but not the secrets people actually want a secret-manager for. The remaining "fetch plaintext config strings from GitHub" use case isn't strong enough to keep the code on life support.

Closing parent issue #17 with the same rationale, and filed #46 to track full removal of the github source backend, the `google/go-github` dependency, and the hidden TUI wizard branch.

The schema-polish + test work in this PR was solid; if the source ever returns from a different design, the test scaffolding here is worth grabbing from the git history.

PR #28, opened 2026-05-07
fix(tui): wire remote-source variables into the mapping form

Galvill:
## Summary

The mapping form's **variables** row only opened `newVarTreeScreen`, which is local-only. AWS- and GitHub-source vars had no entry point — users could configure the source, click *Test* and see "connection OK", and then have nowhere to actually pick a variable.

When at least one non-local source is configured, the variables row now opens a small popup menu:

```
Local secrets…
Add from <source-name>…           (one per remote source)
Back
```

Picking a remote entry launches the existing `var_wizard` (`startVarWizard` → `pickSourceStep` → `dispatchByType` → AWS generic flow / GitHub scope flow) pre-seeded with that source. The wizard's `onComplete`:

- Appends the finished `VarRef` to the mapping (replacing one with the same source/ref/key/name to keep re-runs idempotent).
- `popUntilMsg` back to the mapping form, skipping past the popup, source picker, and any input screens in between.
- Emits `dirtyMsg` + `mappingChangedMsg` so the form's "N selected" count refreshes.

Adds a read-only "Remote variables:" listing under the form rows (using the existing `summariseVar` helper) so the user can verify what they added without diving back into the wizard.

`localVarCount` is updated to count non-local vars too — single-key remote refs as 1, local bag-expands as the bag's key count. The header counter stays accurate.

Single-source local-only configs are unchanged: the popup is skipped and the var-tree opens directly.

## Test plan
- [x] `make test` — green.
- [x] `golangci-lint run ./...` clean.
- [ ] Reviewer: configure an AWS source, hit Test → OK. Open / create a mapping. On the variables row press Enter; the popup should now show "Local secrets…", "Add from <aws-source>…", "Back".
- [ ] Reviewer: walk the wizard (Identifier → optional sub-key → env var name) and confirm landing back on the mapping form with the new var listed under "Remote variables:" and the count incremented.
- [ ] Reviewer: same flow for GitHub source (scope picker → owner/repo → variable name → env name).
- [ ] Reviewer: configure NO remote sources and confirm the variables row still opens the local var-tree directly (no popup).

## Follow-up scope (intentionally not in this PR)
- Editing or removing an existing remote VarRef from the mapping. The form lists them read-only; today removal still requires editing `config.toml` by hand or re-creating the mapping.
- The `//nolint:unused` markers in `var_wizard.go` are now redundant (the wizard is reachable). They're harmless under the current linter set; cleaning them up is a separate cosmetic PR.

### DIFF
--- a/internal/tui/mappings_screen.go
+++ b/internal/tui/mappings_screen.go
@@ -2,6 +2,7 @@ package tui
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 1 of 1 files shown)


Galvill:
Folded into #23 — that PR now includes both the AWS TUI re-enable and the variables-row remote-source entry point.

Issue #30, opened 2026-05-07
bash hook: tab-completion after a mapped command triggers jitenv prompt


Galvill:
## Symptom

After running a mapped file in a bash session, pressing Tab to complete the next command triggers the jitenv hook — including the red 10-second \"agent unreachable\" countdown when the agent is locked. Reproduces whether jitenv is unlocked or not.

## Reproduction

1. Configure a mapping (any file works).
2. In a bash shell with the hook installed (`jitenv hook install`), run that mapped file once.
3. Start typing the next command and press Tab.

Expected: tab-completion runs normally with no jitenv involvement.
Actual: the DEBUG trap fires on commands inside the completion function; on a locked agent the red countdown paints.

## Root cause (working theory)

`internal/shell/snippets/bash.sh` installs a `DEBUG` trap under `extdebug`. Bash runs the DEBUG trap on commands inside programmable-completion functions too. The hook then `stat`s those internal commands, calls `jitenv is-mapped`, and routes through the agent-unreachable warning path on rc=2.

zsh's `preexec` does not fire during completion, so this is bash-only.

## Likely fix

In `__jitenv_debug_trap`, bail out when running inside a completion context. The most reliable signal:

```bash
[[ -n \"\${COMP_LINE-}\" || -n \"\${COMP_POINT-}\" ]] && return 0
```

`COMP_LINE` / `COMP_POINT` are only set by bash while a completion function is executing. Adding this near the top of the trap (after the `__JITENV_REENTRY` short-circuit) should cover the case.

Worth also considering:
- Skip when `\$- != *i*` (non-interactive shells) — defensive.
- Skip when `BASH_COMMAND` matches `compgen`/`complete`/`_*` — belt-and-suspenders if there are completion paths that don't set COMP_LINE.
- Confirm the resolved-path check already rejects directories.

## Test plan after fix

- Lock the agent. Run a mapped script. Type `./mapped-script.sh<Tab><Tab>` and confirm no red countdown appears.
- With the agent unlocked, confirm tab-completion still works and no `jitenv is-mapped` calls fire (set `JITENV_HOOK_DEBUG=1` and verify nothing logs during Tab presses).
- Confirm running the mapped file directly (no Tab) still routes through `jitenv run` as expected.

## References

- `internal/shell/snippets/bash.sh:51` — `__jitenv_debug_trap`.
[...]

PR #31, opened 2026-05-07
feat: cwd_glob mapping kind + bash hook completion fix

### LINKED ISSUE #27: Capture mappings by current working directory (cwd_glob)
## Summary

Add a new mapping kind that matches against the **current working directory** rather than the executable path, so bare-PATH commands (`npm`, `python`, `bundle`, …) run inside a configured directory get env-var injection too.

Today the shell hook only intercepts commands whose first token is an absolute or `./`/`../`-relative path — see `internal/shell/snippets/bash.sh:64-70`. Bare PATH lookups are intentionally skipped:

```bash
if [[ "\$first" = /* ]]; then
    resolved="\$first"
elif [[ "\$first" = ./* || "\$first" = ../* ]]; then
    resolved="\$(cd "\$(dirname "\$first")" 2>/dev/null && pwd)/\$(basename "\$first")"
else
    return 0   # bare PATH lookup → ignored
fi
```

A cwd-keyed mapping closes that gap for the common "I'm working on project X and I want every command I run from inside it to see X's credentials" case.

## Proposed shape

New mapping kind alongside `path` / `glob`:

\`\`\`toml
[[mappings]]
cwd_glob = "~/work/acme/**"
command  = "npm"          # optional — empty means ANY command run in this cwd
[[mappings.vars]]
source = "local"
ref    = "acme_creds"
\`\`\`

Lookup rule: ancestor-walk. \`\$PWD\` matches \`cwd_glob\` if the glob matches \$PWD itself or any descendant. Declaration order still wins on conflicts (consistent with existing globs).

## Implementation sketch

1. **\`config.Mapping\`** — add \`CwdGlob string\` and \`Command string\` fields. \`Config.Validate()\` enforces exactly one of \`Path\` / \`Glob\` / \`CwdGlob\`.
2. **\`config.Index\`** — second lookup table keyed by cwd glob.
3. **Resolver** — new \`IsMappedCwd(pwd, command string) bool\` plus a fetch path that takes (pwd, command) instead of an absolute file path.
4. **Agent protocol** — extend \`is_mapped\` / \`fetch_env\` ops with optional \`cwd\` and \`command\` fields. Existing path-keyed callers stay unchanged (\`is_mapped\` keeps the same exit-code contract: 0 / 1 / 2).
5. **Shell hook** — when the first token is a bare name, ask the agent \`jitenv is-mapped --cwd "\$PWD" --cmd "\$1"\` (currently the hook returns 0 early in that branch). On rc=0, re-exec via \`jitenv run --cwd-mapping --cmd "\$1" -- "\$@"\`.
6. **TUI** — third \"kind\" choice in the mapping form (\"cwd\"), plus a command field rendered when kind=cwd.

## Tradeoffs to think through

- **Hook performance.** Bare PATH commands currently short-circuit before the socket round-trip. Adding cwd matching means *every* command becomes a socket call, which is what the project deliberately avoided. Mitigation: only enable cwd mapping when the user has at least one \`cwd_glob\` mapping configured (the hook can read this from a tiny env var the agent exports, or from \`is-mapped\` returning a fast \"no cwd mappings\" sentinel).
- **Security model widening.** Today secrets only enter a specific exec'd child. With cwd mapping, *every* command run inside the matched directory gets the secrets. That's the feature, but it deserves a security-model section in \`docs/security.md\`.
- **Don't drift toward direnv.** Preserve the invariant: secrets enter the exec'd child's process tree only, never the parent shell. No tempting \"if cwd matches, export into the shell.\"
- **Glob semantics.** Decide whether \`~/work/**\` matches \`~/work/acme/sub/dir\` (probably yes — that's what users expect). The existing \`bmatcuk/doublestar/v4\` matcher handles this; the resolver just needs to walk ancestors.
- **\`.jitenvrc\` per-folder.** A natural extension (per-directory marker file rather than central config) but probably out of scope for v1 — a single encrypted central config keeps the encryption story simple.
- **Subprocess inheritance.** A child process \`cd\`-ing out of the matched cwd doesn't lose the env vars (it inherited them at spawn). Consistent with how shell exports work; document.

## Acceptance criteria

- [ ] \`Config.Mapping\` carries \`cwd_glob\` + \`command\`; \`Config.Validate()\` rejects mappings with multiple kinds set.
- [ ] Agent's \`is_mapped\` and \`fetch_env\` accept \`cwd\` + \`command\` (back-compat for existing callers).
- [ ] Shell hooks (bash + zsh) ask the agent for bare-PATH commands when cwd-glob mappings exist; otherwise behave exactly as today.
- [ ] TUI lets the user create / edit cwd-glob mappings and pick which command(s) to scope to.
- [ ] \`docs/concepts.md\` documents the cwd kind; \`docs/security.md\` notes the model widening.
- [ ] e2e test under \`internal/run\` or \`internal/shell\` covers \"run \\\`npm\\\` from inside a cwd-mapped directory and confirm env vars land\".

## References

- Existing path/glob lookup: \`config.Index\` in \`internal/config/index.go\`.
- Hook trigger semantics: \`internal/shell/snippets/{bash,zsh}.sh\`, exit-code contract documented in \`docs/shell-hook.md\` (after #14 lands).

Galvill:
Combines two issues into one PR.

## #27 — cwd_glob mapping kind

A third mapping kind that matches by current working directory rather than the executed file's path. Bare-PATH commands (`npm`, `python`, `bundle`, …) run inside a configured directory now get env-var injection too. Optional `command` field scopes a mapping to a single bare name; empty matches every command.

```toml
# Anything I run inside ~/work/acme gets the API token.
[[mappings]]
cwd_glob = "~/work/acme"
[[mappings.vars]]
name   = "ACME_API_TOKEN"
source = "local"
ref    = "acme"
key    = "API_TOKEN"

# Only npm in that directory needs the registry token.
[[mappings]]
cwd_glob = "~/work/acme"
command  = "npm"
[[mappings.vars]]
name   = "NPM_TOKEN"
source = "local"
ref    = "acme"
key    = "NPM_TOKEN"
```

### How it's wired
- **Schema**: `config.Mapping` gains `CwdGlob` and `Command` fields. `Validate` enforces exactly one of path/glob/cwd_glob and rejects `command` without `cwd_glob`.
- **Index**: `cwdGlobs` slice with an ancestor-walk match — `cwd_glob = "~/work/acme"` also fires for `~/work/acme/sub/dir`, no trailing `**` needed. Tilde expansion happens at index-build time.
- **Resolver**: new `HasCwdMappings` / `IsMappedCwd` / `FetchEnvCwd`. The fetch-vars loop is shared between the path-keyed and cwd-keyed paths.
- **Agent protocol**: `Request` gains optional `Cwd` + `Command`. Existing path-keyed callers are unchanged; the server dispatches by their presence.
- **Hooks**: bash + zsh check a tiny `has-cwd` sentinel file the agent maintains in its runtime dir; if the file isn't there (no cwd_glob mappings configured) the bare-PATH branch short-circuits, so users who don't use this feature pay zero overhead. With the sentinel present, the hook asks the agent and re-execs through `jitenv run --cwd $PWD --cmd <name>` on rc=0. rc=1 / rc=2 silently fall through — bare commands fire on every keystroke-completed name, so warning on every one would be too noisy.
- **CLI**: `jitenv is-mapped --cwd <pwd> --cmd <name>` and `jitenv run --cwd <pwd> --cmd <name> [args…]`. Run resolves the bare name on `$PATH`, fetches env by (cwd, command), and exec's with `argv[0]` preserved as the typed name.
- **TUI**: kind picker gets a third "cwd" choice. When a mapping is in cwd mode, the form grows a "command" row above "variables".
- **Docs**: `docs/concepts.md` documents the new kind, ancestor-walk match, and sentinel perf model. `docs/security.md` calls out the trust-model widening — every command in the matched cwd gets the secrets, not just the one explicit file.

### Acceptance criteria from the issue
- [x] `Config.Mapping` carries `cwd_glob` + `command`; Validate rejects multi-kind mappings.
- [x] Agent's `is_mapped` and `fetch_env` accept `cwd` + `command` (back-compat for existing callers).
- [x] Shell hooks ask the agent for bare-PATH commands when cwd-glob mappings exist; otherwise behave exactly as today.
- [x] TUI lets the user create / edit cwd-glob mappings and pick the command.
- [x] `docs/concepts.md` documents the cwd kind; `docs/security.md` notes the model widening.
- [x] e2e test (`internal/shell/hook_cwd_test.go`) covers "run a bare command from inside a cwd-mapped directory and confirm env vars land".

## #30 — bash hook tab-completion fix

`__jitenv_debug_trap` now bails out at the top when `COMP_LINE` or `COMP_POINT` is set. Bash runs the DEBUG trap on commands inside programmable-completion functions too, which is what was painting the red "agent unreachable" countdown when the user pressed Tab after running a mapped file with a locked agent. The COMP_* vars are bash's own signal that the trap is firing inside a compfunc, so the guard is exactly the right shape.

I considered also bailing on non-interactive shells (`[[ "$-" != *i* ]]`) but reverted it because the e2e test suite drives the hook from `bash -c '…'`, which IS non-interactive. The COMP_* guard alone is enough; non-interactive shells don't run completion.

## Test plan
- [x] `make test` — green, including the new mapping_test.go (8 cases) and hook_cwd_test.go end-to-end.
- [x] `golangci-lint run ./...` — clean.
- [x] e2e: agent up + bash hook + cwd_glob mapping + bare command → FOO arrives.
- [ ] Reviewer (interactive bash): unlock agent, run a mapped script once, lock the agent, then `./mapped<TAB><TAB>` — confirm no red countdown appears. With `JITENV_HOOK_DEBUG=1` set, no `is-mapped` calls should log during Tab presses.
- [ ] Reviewer: confirm running the mapped file directly (no Tab) still routes through `jitenv run`.
- [ ] Reviewer (TUI): create a cwd_glob mapping, pick a command, save. Confirm `cat config.toml` round-trips correctly and that the `has-cwd` sentinel appears in `\$XDG_RUNTIME_DIR/jitenv/`.

## Notes
- Hook performance for users who don't use cwd_glob is still zero socket calls per command (the sentinel file isn't there, so the stat alone fails in microseconds).
- `jitenv run --cwd ... --cmd ...` uses `DisableFlagParsing` and a hand-rolled splitter so `--anything` arguments to the wrapped command pass through cobra unmolested.

Closes #27 and #30.

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -65,19 +65,59 @@ The "expand" shape is the one most people miss reading the schema.
 It's exactly what the TUI's "tick the bag" mode produces — useful when
 a bag's keys already match the env var names you want.
 
-## Path vs. glob
+## Mapping kinds: path, glob, cwd_glob
 
-A mapping's target is either a `path` (string, absolute or
-`~`-relative) or a `glob` (matched with [doublestar](https://github.com/bmatcuk/doublestar/v4)).
+A mapping picks **one** target shape:
 
-- Path matches are exact and faster.
-- Glob patterns support `**` (recursive), `*`, `?`, `[…]` and curly
-  alternation. Common useful patterns: `~/work/**/*.sh`,
-  `**/scripts/deploy*`.
-- Lookup order is **declaration order**: exact paths first, then each
-  matching glob. When two entries provide the same env var name, the
-  later one wins. This is intentional — a glob can supply defaults
-  while a more specific path overrides individual vars.
+- `path` — exact filesystem path. Fast lookup.
+- `glob` — [doublestar](https://github.com/bmatcuk/doublestar/v4)
+  pattern matched against the executed file's resolved path. Supports
+  `**`, `*`, `?`, `[…]` and curly alternation. Common patterns:
+  `~/work/**/*.sh`, `**/scripts/deploy*`.
+- `cwd_glob` — pattern matched against `$PWD` (and any ancestor). For
+  cwd mappings the shell hook intercepts **bare-PATH commands**
+  (`npm`, `python`, `bundle`, …) run from inside the matched directory
+  — useful for "anything I run inside this project gets the project's
+  env vars". An optional `command = "<bare-name>"` field scopes the
+  mapping to a single command; empty means every command run inside
+  the matched cwd.
+
+Lookup order is **declaration order** within each kind: exact paths
+first, then each matching glob, then each matching cwd_glob. When two
+entries provide the same env var name, the later one wins — so a
+glob can supply defaults while a more specific path overrides
+individual vars.
+
+```toml
+# Anything run inside ~/work/acme gets ACME_API_TOKEN.
+[[mappings]]
+cwd_glob = "~/work/acme"
+[[mappings.vars]]
+name   = "ACME_API_TOKEN"
+source = "local"
+ref    = "acme"
+key    = "API_TOKEN"
+
+# Only npm in that directory needs the registry token.
+[[mappings]]
+cwd_glob = "~/work/acme"
+command  = "npm"
+[[mappings.vars]]
+name   = "NPM_TOKEN"
+source = "local"
+ref    = "acme"
+key    = "NPM_TOKEN"
+```
+
+### Performance: the has-cwd sentinel
+
+The bash and zsh hooks short-circuit on bare-PATH commands when no
+cwd_glob mapping is configured, so users who never use cwd mappings
+pay zero overhead. The agent maintains a tiny sentinel file at
+`$XDG_RUNTIME_DIR/jitenv/has-cwd` (or `$TMPDIR/jitenv-<uid>/has-cwd`)
+when at least one cwd_glob mapping exists; the hook checks for the
+file's existence (one stat) before spending a socket round-trip. The
+sentinel is created on agent start and refreshed on every reload.
 
 ## Agent
 
--- a/docs/security.md
+++ b/docs/security.md
@@ -118,6 +118,45 @@ Atomic save via `config.AtomicSave` (sibling tempfile + rename, mode
   fiddling and offers little against the local-root threat that
   actually matters.
 
+## cwd_glob mappings widen the blast radius
+
+Path and glob mappings only inject env vars into the *one specific
+file* you executed — same one-process boundary as `jitenv run`. A
+cwd_glob mapping is broader: **every command run inside the matched
+directory** that matches the optional `command` filter gets the env
+vars. The boundary is now "one descendant process at a time", but
+the set of triggering commands is anything you (or anything you run)
+types in that directory.
+
+What's still protected:
+
+- The parent shell never has the secrets in its own environ. Each
+  triggering command gets them at exec, and they live in that
+  child's process tree only.
+- `cd`-ing out of the mapped directory in the calling shell does
+  **not** strip the secrets from already-running children — they
+  inherited them at spawn time, exactly like manual exports work.
+- The hook's `command` filter (when set) restricts firing to that
+  bare name. `command = "npm"` will not fire for `git`, `ls`, or
+  anything else even inside the matched cwd.
+
+What to be aware of:
+
+- Inside a matched cwd, **any** command (or the optional
+  `command`-scoped one) gets the secrets. If a malicious binary on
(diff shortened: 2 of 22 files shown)


Galvill:
Superseded by the wrapper-based design in #33 (docs/design/wrapper-hook.md). The schema/protocol bits get re-cut in the implementation PR with the design's renames (Command → Commands []string).

PR #33, opened 2026-05-08
docs(design): split-hook architecture for cwd_glob mappings

Galvill:
Captures the brainstorm conversation + the decisions you made.

## What's in the doc
- **The split**: path-mapped via DEBUG/preexec trap (current, works), cwd-mapped via PATH-prepend wrappers (new).
- **Shim design**: `jitenv __shim` subcommand, dispatched off `os.Args[0]` basename, resolves the real command via $PATH minus its own dir, then `syscall.Exec`s with merged env.
- **Per-shell wrapper dir**: `\$XDG_RUNTIME_DIR/jitenv/shells/<pid>/bin/`, populated on `chpwd`, GC'd by the agent on next start.
- **Hook integration**: bash `PROMPT_COMMAND` PWD-diff, zsh native `chpwd_functions` — both call `jitenv __chpwd <pid> <oldpwd> <newpwd>`.
- **Migration plan** vs PR #31: reuses schema/protocol with one breaking rename (`Command` → `Commands []string`), drops the bare-PATH branch from the shell hooks.

## Decisions (you said)
1. Explicit `commands = [...]` only — wildcard rejected as a footgun.
2. `jitenv __shim` subcommand, not a separate binary.
3. GC on next agent start (no per-shell EXIT trap); `XDG_RUNTIME_DIR` wipe at logout is the safety net.
4. Same code path for bash + zsh.
5. Direnv coexistence deferred.

## What this PR is NOT
Code. This is just the design doc. Implementation lands as a separate PR (likely re-purposing or replacing #31).

## Test plan
- [x] Doc-only change; `make test` unaffected.
- [ ] Reviewer: skim the architecture section against your mental model and flag anything I got wrong before we cut code.

### DIFF
--- a/docs/design/wrapper-hook.md
+++ b/docs/design/wrapper-hook.md
@@ -1,0 +1,222 @@
+# Design: split the hook — DEBUG trap for paths, PATH-prepend wrappers for cwds
+
+Status: **Approved direction**. Supersedes the all-DEBUG-trap cwd_glob
+implementation in PR #31; the schema and agent-protocol changes from that
+PR mostly carry over (one rename — see [Migration](#migration)).
+
+## Problem
+
+The current shell hook uses bash's DEBUG trap (zsh: `preexec`) to do two
+unrelated jobs:
+
+1. **Path-mapped scripts** — intercept `./deploy.sh`, `~/scripts/foo`, etc.,
+   re-exec through `jitenv run` with merged env.
+2. **Cwd-mapped commands** — when running any command inside a configured
+   directory, inject env vars.
+
+Job 1 is a good fit for the DEBUG trap: `BASH_COMMAND` is unambiguous for an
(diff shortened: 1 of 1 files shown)


Galvill:
Closing as superseded. The design described here was implemented and shipped in #34 (released as v0.4.0). The doc was framed as a pre-implementation "approved direction" with migration notes from PR #31 — landing it after the fact would just bake stale framing into the repo. The architecture is now captured in:

- The #34 PR description (cwd_glob, wrapper dir, shim, chpwd op).
- Commit messages on `feat/cwd-wrapper` (`cdd4691 feat: cwd_glob mappings via PATH-prepend wrappers`, `240ad68 fix(chpwd): read config directly`, etc.).
- `CLAUDE.md` and the `internal/shim` / hook code itself.

If we want a permanent design doc, it should be a fresh write-up reflecting shipped reality, not this branch.

Issue #91, opened 2026-05-13
#39 stage 2.6: Windows config-dir, GoReleaser windows targets, CI elevation


Galvill:
Sub-issue 6 of 6 for #39 Stage 2+. Last to land; gates the v1 Windows release.

## Scope

Three small pieces that ship the port end-to-end:

1. **Config-dir resolution** — Split `internal/config/path.go` so `DefaultPath` returns `%APPDATA%\jitenv\config.toml` on Windows (currently Unix-only — `XDG_CONFIG_HOME` then `~/.config`).
2. **GoReleaser** — Add `windows/amd64` + `windows/arm64` to `.goreleaser.yaml`'s `goos` list and a `.zip` archive format for windows artifacts.
3. **CI elevation** — Flip the `test-windows` CI job from build-only to running the previously-excluded agent / run / shim / shell suites (drop the `//go:build !windows` guards on tests that should now pass after stages 2.1-2.5 land).

## Decisions (resolved up front)

- **Windows packaging scope**: `.zip` only for v1. Defer winget, Chocolatey, Scoop, MSI to follow-up issues. Reasoning: ships the v1 port without external account/repo dependencies; the four package-manager paths are independent CI flows and can land incrementally.

## Touches

- `internal/config/path.go` (split into `_unix.go` / `_windows.go`)
- `.goreleaser.yaml`
- `.github/workflows/ci.yml`
- Build tags on `internal/agent/daemonize_test.go`, `internal/run/run_e2e_test.go`, `internal/shell/hook_*_test.go` (drop `!windows` where the test now passes)

## Blocks / blocked-by

- **Blocked-by**: stages 2.1–2.5.
- **Blocks**: nothing in this chain. Optional follow-ups (winget, Chocolatey, signing) are separate issues.

## Acceptance

- [ ] `go test ./...` on `windows-latest` runs the agent / run / shim suites and passes.
- [ ] `goreleaser release --snapshot` produces `jitenv_<v>_windows_amd64.zip` (and `_arm64`) containing the binary + `snippets/powershell.ps1` + LICENSE + README.
- [ ] A fresh Windows VM can: download the zip, run `jitenv hook install`, `jitenv unlock`, `cd` into a cwd_glob dir, run a wrapped command, see env vars injected.
- [ ] README's platform-support section flips from "Windows: not yet supported" to "Windows: supported via cwd_glob (PowerShell 7+)".
- [ ] CLAUDE.md's platform note updated.

## Out of scope (separate issues)

[...]

Issue #89, opened 2026-05-13
#39 stage 2.4: .ps1 wrapper shims + Windows-aware chpwd reconcile


Galvill:
Sub-issue 4 of 6 for #39 Stage 2+.

## Scope

Currently `internal/chpwd/chpwd.go` calls `os.Symlink(target, link)` to populate the per-shell wrap dir. On Windows replace this with writing tiny `.ps1` files (one per wanted command) of the shape `& jitenv __shim <name> @args` — see the snippet in issue #39. Make the reconcile platform-aware (extension `.ps1` instead of bare name, content-based equality instead of `Readlink`). Update `ShellWrapDir` callers and the shim's `excludeDir` self-detection (`internal/shim/shim.go:~176`) to handle the new file shape.

## Touches

- `internal/chpwd/chpwd.go` (split `reconcile` into `_unix.go` / `_windows.go` or branch internally)
- `internal/shim/shim.go` (`lookPathExcluding` already handles `\` correctly via `os.PathSeparator`; verify)
- Possibly `internal/agent/paths.go` for wrap-dir layout

## Blocks / blocked-by

- **Blocked-by**: stage 2.3 (shim must actually run before there's anything to wrap).
- **Blocks**: stage 2.5.

## Acceptance

- [ ] On Windows, `jitenv __chpwd` populates `%LOCALAPPDATA%\jitenv\shells\<pid>\bin\` with `.ps1` files for each matched command.
- [ ] Re-running with the same desired set is a no-op (content hash matches).
- [ ] Unix symlink behaviour and tests unchanged.
- [ ] `GcOrphanShells` reaps the Windows wrap dirs the same way it does on Unix.

## Out of scope

- `.exe` shim files (would require shipping a per-command tiny launcher binary). `.ps1` is enough for the cwd_glob v1 flow.

Issue #87, opened 2026-05-13
#39 stage 2.2: hidden-console daemon spawn + key-handle handoff


Galvill:
Sub-issue 2 of 6 for #39 Stage 2+.

## Scope

Implement `SpawnDaemon` in `daemonize_windows.go` using `cmd.Start()` with `CREATE_NO_WINDOW` and pass the master key to the child via an inherited `*os.File` handle (mirrors `ExtraFiles` fd 3 on Unix; on Windows requires `SysProcAttr.AdditionalInheritedHandles` and a child-side flag like `--key-handle=<hex>` instead of `--key-fd=3`). Update the `__agent` subcommand argument parsing to accept the Windows form. Lock = signal child to exit (named event or pipe-close); pidfile semantics stay the same.

## Touches

- `internal/agent/daemonize_windows.go`
- `cmd/jitenv` / `internal/cli/agent_internal.go` (key-fd vs key-handle flag)
- `internal/agent/pidfile.go` if `PidAlive` needs a Windows path
- `internal/cli/unlock.go` (remove the `runtime.GOOS == "windows"` guard at the top)

## Blocks / blocked-by

- **Blocked-by**: #stage-2-1 (named-pipe transport — needs a working transport for the "socket present" readiness check).
- **Blocks**: stages 2.3, 2.5, 2.6.

## Acceptance

- [ ] `jitenv unlock` on Windows starts a detached agent; the parent process returns once the pipe is reachable.
- [ ] `jitenv status` from a new shell hits the running agent and gets a real `Status` response.
- [ ] `jitenv lock` cleanly shuts the agent down; pidfile + pipe gone.
- [ ] The `runtime.GOOS == "windows"` guard at the top of `internal/cli/unlock.go` is removed.
- [ ] The master key is not visible in the child's command line, environment variables, or any file on disk.

## Open question for implementer

Child shutdown mechanism — named Windows event, pipe-close, or `os.Process.Signal(os.Interrupt)`? `os.Interrupt` on Windows is effectively a kill; named events are most idiomatic but require extra wiring. Default to pipe-close (when the agent's listener closes, the loop exits) and add a named event only if pipe-close turns out to be racy.

## Out of scope

- Service-host (svchost) mode. Hidden-console subprocess is the v1 model.

Issue #39, opened 2026-05-09
Add native Windows support with PowerShell hook


Galvill:
## Summary

Ship native Windows support so PowerShell users can install jitenv without going through WSL. This is a **design + tracking** issue — the deliverable is a written design doc and a go/no-go decision, similar to #13 (macOS) and #18 (keychain).

WSL already works today (it's Linux); this issue is specifically about a Windows-native binary that runs alongside `pwsh.exe`.

## Decided up-front

- **PowerShell 7+ only.** No support for the legacy Windows PowerShell 5.1 — it's a separately-installed shell on modern Windows, the audience overlap with jitenv's "I want a real secrets workflow" user is minimal, and PS 7's PSReadLine + module ergonomics are meaningfully nicer to write against. Users who only have 5.1 can install PS 7 from `winget install Microsoft.PowerShell`.
- **`.ps1` shims for the cwd_glob wrapper directory.** Runs in-process inside the calling pwsh, no `cmd.exe` spawn cost per command. The trade-off (only PowerShell callers see the wrapper, not `cmd.exe` invocations from elsewhere) is acceptable: jitenv's hook is registered in the user's `pwsh` profile, so the wrapper is for pwsh-originated commands by definition.

## Scope of the port

The current agent is Linux- (and now darwin-) only. Several things have no Windows equivalent and need a different design:

| Concern | Linux/macOS | Windows |
|---|---|---|
| IPC transport | Unix domain socket (`agent.sock`, mode 0600) | named pipe (`\\.\pipe\jitenv-<sid>`) — no FS perms; ACL via SDDL |
| Peer credential check | `SO_PEERCRED` / `LOCAL_PEERCRED` | `GetNamedPipeClientProcessId` + token SID compare |
| Daemon model | Setsid double-fork, `XDG_RUNTIME_DIR/jitenv/agent.sock` | Windows Service? tray app? hidden console child? |
| Runtime dir | `$XDG_RUNTIME_DIR` / `$TMPDIR` | `%LOCALAPPDATA%\jitenv` |
| Config dir | `$XDG_CONFIG_HOME` / `~/.config` | `%APPDATA%\jitenv` |
| Cwd_glob shims | symlinks to the `jitenv` binary | `.ps1` shims (decided above) |
| Shell hook trigger | bash `DEBUG` trap, zsh `preexec` | pwsh `prompt` function override + `Set-PSReadLineKeyHandler` for completion-time guards |

## PowerShell 7 hook shape

[...]

PR #100, opened 2026-05-13
feat(shell,hook): PowerShell 7 hook + jitenv hook powershell + installer (#90)

### LINKED ISSUE #90: #39 stage 2.5: PowerShell 7 hook snippet + jitenv hook powershell + installer
Sub-issue 5 of 6 for #39 Stage 2+.

## Scope

Add `internal/shell/snippets/powershell.ps1` embedded via `//go:embed`, a `pwsh` case in `Render`, and a `powershell` subcommand in `internal/cli/hook.go`. The snippet is the `prompt` override from issue #39 (chpwd-on-prompt). **v1 ships cwd_glob only** — no `Set-PSReadLineKeyHandler` path-mapped flow. Extend `internal/shell/install.go` with a Windows install path that appends the eval block to `$PROFILE` (`Documents\PowerShell\Microsoft.PowerShell_profile.ps1`); `DetectShell` needs a Windows branch since `$SHELL` is unset there. Update `warnIfHookMissing` accordingly.

## Touches

- `internal/shell/snippets/powershell.ps1` (new)
- `internal/shell/snippets.go`
- `internal/shell/render.go`
- `internal/shell/install.go`
- `internal/cli/hook.go`
- `internal/cli/unlock.go` (status text)

## Blocks / blocked-by

- **Blocked-by**: stage 2.2 (agent) and 2.4 (chpwd wrap dir).
- **Blocks**: nothing — this is the final user-facing piece for the runtime port.

## Acceptance

- [ ] `jitenv hook powershell` prints a self-contained `.ps1` snippet with `{{RuntimeDir}}` / `{{ConfigPath}}` markers substituted (matches the bash/zsh hook contract).
- [ ] `jitenv hook install` from a pwsh session edits `$PROFILE` idempotently.
- [ ] `jitenv hook status` reports the pwsh state correctly.
- [ ] Manual smoke test: fresh pwsh → `cd` into a cwd_glob dir → `npm.ps1` shim runs → env injected.

## Out of scope

- **No** PSReadLine `AcceptLine` path-mapped flow (the equivalent of the bash `DEBUG` trap intercepting absolute paths). Decided per issue #39 — cwd_glob only for v1.
- Legacy PowerShell 5.x support. pwsh 7+ only.
- `ConPTY` vs legacy-console quirks for the agentwarn countdown — assume modern terminal; revisit if user reports break.

Galvill:
## Summary

Final user-facing piece of the #39 Windows port chain. Adds the PowerShell 7+
hook snippet, the `jitenv hook powershell` cobra subcommand, and the
`$PROFILE`-pointed installer. Linux/macOS behaviour is untouched; only #91
(release packaging) remains.

### Why override `prompt` and not PSReadLine?

PowerShell has no DEBUG trap or preexec hook. The `prompt` function is the
one place pwsh guarantees to call once per interactive submission, so it's
the natural seam for the chpwd-on-prompt half of the flow. A PSReadLine
`AcceptLine` handler would let us intercept absolute-path commands the way
bash's DEBUG trap does, but #39 rejected that as too invasive — PSReadLine
isn't always loaded (every $PROFILE that hides the module banner with
`Set-PSReadLineOption` first has to load it), and `Set-PSReadLineKeyHandler`
fights tooling like VS Code's integrated terminal. v1 ships cwd_glob only;
users on Windows route through the `.ps1` wrappers populated by #89.

The snippet captures the user's existing `prompt` as a scriptblock
(`$function:prompt`) before overriding it, then chains: `__jitenv_chpwd`
runs first, then the original prompt renders. Errors are trapped so a
chpwd failure never wedges the prompt; debug output is gated behind
`$env:JITENV_HOOK_DEBUG` (matches the bash/zsh hook).

### `$PROFILE` install behaviour

`jitenv hook install` (or `--shell powershell`) appends an
`Invoke-Expression (& jitenv hook powershell | Out-String)` block to
`$PROFILE.CurrentUserCurrentHost`. The installer queries pwsh itself
(`pwsh -NoProfile -Command '$PROFILE.CurrentUserCurrentHost'`) so users
with OneDrive-redirected Documents folders or custom host configurations
end up touching the right file; if pwsh isn't on PATH at install time we
fall back to the documented default —
`%USERPROFILE%\Documents\PowerShell\Microsoft.PowerShell_profile.ps1` on
Windows, `~/.config/powershell/Microsoft.PowerShell_profile.ps1` on Unix.
Idempotency uses the same exact-line check as bash/zsh; re-running
`jitenv hook install` is a no-op once the line is present.

PowerShell has no analogue of the bash login chain (the bash
`.bash_profile` → `.bash_login` → `.profile` walk), because pwsh sources
`$PROFILE` for every interactive host. `InstallReport.LoginPath` /
`LoginAdded` therefore stay empty for the pwsh path, matching the zsh
behaviour.

### `DetectShell` on Windows

`$SHELL` is normally unset on Windows. `runtime.GOOS == "windows"` short-
circuits `DetectShell` to return `"powershell"` unconditionally — cmd.exe
and PowerShell 5.x (the `WindowsPowerShell` 5.x profile path) are
explicitly unsupported per the #39 design decision. Users who want a
different shell can still pass `--shell <name>`; auto-detection just
defaults to the only supported one.

### Snippet quoting

`pwshQuote` doubles embedded single quotes (`don''t`) instead of bash's
`'\''` close-escape-reopen, and leaves backslashes alone so Windows paths
like `C:\Users\Alice\AppData\Local\jitenv` round-trip unmodified. The
template uses `{{RuntimeDir}}` / `{{ConfigPath}}` markers identical to
bash/zsh; `Render` dispatches to the appropriate quoter.

## Test coverage

- `internal/shell/render_test.go` — Render(`powershell`/`pwsh`) bakes paths,
  emits the wrap-dir + prompt-override plumbing, and pwshQuote round-trips
  Windows paths and escapes embedded single quotes. Runs on all platforms.
- `internal/shell/install_test.go` — `HookLine("powershell")` pins the
  Invoke-Expression line; `InstallShell("powershell")` writes the rc file
  idempotently and leaves LoginPath empty. Unix-side test that exercises
  the documented fallback (PATH set to an empty dir to skip the pwsh probe).
- `internal/shell/install_windows_test.go` (new, `//go:build windows`) —
  `DetectShell` returns `"powershell"`; `RcPath`/`InstallShell`/`CurrentStatus`
  end-to-end under a tempdir-rooted `$USERPROFILE`.
- `internal/cli/hook_test.go` (new) — `jitenv hook powershell` and the
  `pwsh` alias both emit a non-empty snippet with the wrap-dir +
  prompt-override plumbing intact.

Build/vet: `make build`, `make build-windows`, and `GOOS=windows GOARCH=amd64
go vet ./...` are all clean. `go test ./... -count=1` is green on Linux.

## Manual smoke test

Not run — this branch was developed on Linux and the dev box has no pwsh
available. Recommended follow-up:

```powershell
# In a fresh pwsh 7+ session on Windows, against a built jitenv.exe + an
# unlocked config with a cwd_glob mapping for the test directory:
Invoke-Expression (& jitenv hook powershell | Out-String)
cd C:\path\to\mapped\dir
# Expect: __JITENV_WRAP_DIR populated with npm.ps1 etc; running `npm`
# resolves to the wrapper and the agent injects the mapped env vars.
```

## Test plan

- [ ] CI green on linux + macos + windows
- [ ] Manual smoke test in a real Windows pwsh 7+ session (see above)

## Links

- Closes #90
- Predecessor: #89 (.ps1 wrapper shims + Windows-aware chpwd reconcile)
- Predecessor: #87 (named-pipe transport)
- Umbrella: #39 (Windows port)

### DIFF
--- a/internal/cli/hook.go
+++ b/internal/cli/hook.go
@@ -39,6 +39,28 @@ func newHookCmd() *cobra.Command {
 			return nil
 		},
 	})
+	c.AddCommand(&cobra.Command{
+		Use:     "powershell",
+		Aliases: []string{"pwsh"},
+		Short:   "Print PowerShell 7+ integration (Windows)",
+		Long: `Print the PowerShell hook snippet. Source it with:
+    Invoke-Expression (& jitenv hook powershell | Out-String)
+
+The snippet wraps the prompt function to drive cwd_glob reconciliation
+on every prompt and prepends the per-shell wrap dir to $env:Path so
+.ps1 shims resolve via PATHEXT. Absolute-path command interception
+(the bash DEBUG trap equivalent) is intentionally not implemented on
+PowerShell — see issue #39. PowerShell 5.x and cmd.exe are unsupported.`,
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			out, err := shell.Render("powershell")
+			if err != nil {
+				return err
+			}
+			fmt.Fprint(cmd.OutOrStdout(), out)
+			return nil
+		},
+	})
 	c.AddCommand(newHookInstallCmd())
 	c.AddCommand(newHookStatusCmd())
 	return c
@@ -53,7 +75,10 @@ func newHookInstallCmd() *cobra.Command {
 login chain (.bash_profile / .bash_login / .profile) doesn't already
 end up sourcing ~/.bashrc, a guarded source line is added so that
 login shells pick up the hook too. For zsh, the eval line is
-appended to ~/.zshrc (sourced for both interactive and login).
+appended to ~/.zshrc (sourced for both interactive and login). For
+PowerShell, the Invoke-Expression line is appended to
+$PROFILE.CurrentUserCurrentHost (typically Documents\PowerShell\
+Microsoft.PowerShell_profile.ps1 on Windows).
 Re-running this command is a safe no-op when nothing needs to change.`,
 		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, _ []string) error {
@@ -62,7 +87,7 @@ Re-running this command is a safe no-op when nothing needs to change.`,
 				sh = shell.DetectShell()
 			}
 			if sh == "" {
-				return fmt.Errorf("could not detect shell from $SHELL; pass --shell bash|zsh")
+				return fmt.Errorf("could not detect shell from $SHELL; pass --shell bash|zsh|powershell")
 			}
 			rep, err := shell.InstallShell(sh)
 			if err != nil {
@@ -86,7 +111,7 @@ Re-running this command is a safe no-op when nothing needs to change.`,
 			return nil
 		},
 	}
-	cmd.Flags().StringVar(&shellFlag, "shell", "", "shell to install for (bash|zsh); auto-detect by default")
+	cmd.Flags().StringVar(&shellFlag, "shell", "", "shell to install for (bash|zsh|powershell); auto-detect by default")
 	return cmd
 }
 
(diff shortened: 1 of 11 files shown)


Galvill:
Superseded by #101, which bundles this PR's commit verbatim (40ac2cf) plus the Windows passphrase-prompt fix. Merging #101 will land both fixes in a single shot.

PR #137, opened 2026-05-14
fix(agent): defer pr.Close in SpawnDaemon so the key-pipe read end can't leak (#115)

### LINKED ISSUE #115: security: SpawnDaemon (Unix) leaks the key-pipe read end on early-return error paths
## Summary
`internal/agent/daemonize_unix.go:30-49`:

```go
pr, pw, err := os.Pipe()
if err != nil { return err }
defer pw.Close()                       // pw is deferred, pr is NOT

exe, err := os.Executable()
if err != nil { return err }           // pr leaked

logF, err := os.OpenFile(paths.LogFile, ...)
if err != nil { return err }           // pr leaked
defer logF.Close()

devNull, err := os.OpenFile(os.DevNull, ...)
if err != nil { return err }           // pr leaked
defer devNull.Close()

...
if err := cmd.Start(); err != nil {
    pr.Close()
    return err
}
pr.Close()
```

The read end of the master-key pipe is closed only on the success branch and the `cmd.Start()` failure branch — three earlier error returns (`os.Executable`, `os.OpenFile(LogFile)`, `os.OpenFile(devNull)`) leak `pr` as an open file descriptor in the parent.

The Windows version (`daemonize_windows.go`) correctly closes `pr` on each early return, so this is Unix-only.

## Impact
1. **FD leak**: long-lived parent processes (e.g. the TUI calling `SpawnDaemon` after a config save) accumulate stale pipe fds across retried unlocks.
2. **Inheritance risk**: any subsequent `exec.Command` invocation made by the same parent that uses `ExtraFiles` could inadvertently inherit the leaked fd into an unrelated child. (Go's stdlib `exec.Cmd` does explicitly close non-`ExtraFiles` fds via `closeOnExec`, so this is defence-in-depth: the leak is real, the inheritance is not, but the parent-side discipline is broken.)
3. **The pipe is empty at the time of the leak** — the key bytes haven't been written yet, so this is not a direct disclosure. It's a hygiene gap that becomes load-bearing the next time someone adds an `ExtraFiles` call between `os.Pipe()` and the eventual `pr.Close()`.

## Fix
Move the close to a defer immediately after `os.Pipe()`:

```go
pr, pw, err := os.Pipe()
if err != nil { return err }
defer pr.Close()
defer pw.Close()
...
// remove the manual pr.Close() calls at lines 65 and 68
```

A `defer` that runs after `cmd.Start()` succeeded is a no-op on the now-closed fd, but explicit is cleaner: keep one manual `pr.Close()` after `cmd.Start()` returns success and remove the defer.

Location: `internal/agent/daemonize_unix.go:30-68`.

## Severity
**High** for hygiene of a critical code path; not directly exploitable today.

Galvill:
## Summary
Closes #115.

Three early-return paths between `os.Pipe()` and `cmd.Start()` (the `os.Executable` lookup, the log-file `OpenFile`, and the `/dev/null` `OpenFile`) left `pr` — the master-key pipe's read end — open in the parent process. The Windows version already handles this; Unix did not.

## Fix
Add a `defer pr.Close()` immediately after `os.Pipe`, matching the discipline already used for `pw`. The eager success-path close stays so the parent's fd table remains tight while the agent runs.

## Regression test
`TestSpawnDaemonDoesNotLeakPipeFdOnError` (Linux-tagged) polls `/proc/self/fd` before/after a `SpawnDaemon` call against a bad log path. On the old code: `leaked 1 fd(s)`. On the fix: zero leak.

## Test plan
- [x] New test fails on `main`, passes on this branch.
- [x] `go test ./...` clean on Linux.
- [x] `GOOS=windows go vet ./...` clean (Windows path unchanged).
- [ ] CI passes.

### DIFF
--- a/internal/agent/daemonize_unix.go
+++ b/internal/agent/daemonize_unix.go
@@ -31,6 +31,12 @@ func SpawnDaemon(paths Paths, configFile string, idle time.Duration, key []byte)
 	if err != nil {
 		return err
 	}
+	// Both ends are closed on every early-return path; the success path
+	// explicitly closes pr after cmd.Start so the child holds the only
+	// reference, and pw after writing the key so the child's read EOFs.
+	// The defers are safety nets for the OpenFile/Executable error
+	// returns below — without them, pr would leak (security #115).
+	defer pr.Close()
 	defer pw.Close()
 
 	exe, err := os.Executable()
@@ -62,10 +68,12 @@ func SpawnDaemon(paths Paths, configFile string, idle time.Duration, key []byte)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setsid: true}
 
 	if err := cmd.Start(); err != nil {
-		pr.Close()
 		return err
 	}
-	pr.Close() // child holds its own copy
+	// Close eagerly so the child holds the only ref to the read end —
+	// the defer above would still fire on return, but eager close keeps
+	// the parent's fd table tight while the agent runs.
+	pr.Close()
 
 	if _, err := pw.Write(key); err != nil {
 		return fmt.Errorf("write key to child: %w", err)
--- a/internal/agent/daemonize_unix_fd_test.go
+++ b/internal/agent/daemonize_unix_fd_test.go
@@ -1,0 +1,55 @@
+//go:build linux
+
+package agent
+
+import (
+	"os"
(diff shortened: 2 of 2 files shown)


Galvill:

PR #138, opened 2026-05-14
fix(agent): cap concurrent connections + set deadline before peer check (#114)

### LINKED ISSUE #114: security: agent has no concurrent-connection limit and writes 'unauthorized' before setting a deadline
## Summary
Two related issues in `internal/agent/server.go`:

### a) Write to unauthorised peer before any deadline
`server.go:175-177`:
```go
defer c.Close()
if err := checkPeerUid(c); err != nil {
    _ = WriteMessage(c, Response{OK: false, Error: "unauthorized"})  // no deadline yet
    return
}
if err := c.SetDeadline(time.Now().Add(10 * time.Second)); err != nil { ... }
```
The 10-second deadline is only applied after a successful `checkPeerUid`. The `WriteMessage` on the rejection branch has no timeout — a client that connects and never reads holds the goroutine in `c.Write` until the OS send buffer fills, indefinitely.

### b) Unbounded accept loop
`server.go:122-131`:
```go
for {
    conn, err := a.listener.Accept()
    ...
    go a.handle(ctx, conn)
}
```
No semaphore, no `WaitGroup`-with-cap, no rate limit. A misbehaving same-user process can open thousands of connections and exhaust the agent's FD table and goroutine count.

## Impact
A same-user process (the only class that can connect, per the 0600/SDDL gate) can:
1. Hold the agent's goroutines blocked indefinitely on `WriteMessage` to a slow-reading peer **before** authentication.
2. Exhaust the agent's open-file-descriptor limit, causing every legitimate `jitenv run` invocation to fail with `connect: too many open files` → user runs commands without secrets injected.
3. Suppress idle shutdown: `lastSeen` is only updated after a successful peer check (`server.go:190`), so unauthenticated floods don't keep the agent alive — good — but they do consume resources while they last.

Combined, this is a local DoS against the secret-injection path. Not catastrophic, but trivial to trigger by accident (a buggy script) or on purpose.

## Fix
- Move `c.SetDeadline(time.Now().Add(2 * time.Second))` to the **first** line of `handle`, before `checkPeerUid`.
- Bound concurrent connections with a buffered semaphore channel (e.g. cap 64). Drop excess connections immediately after accept:
```go
sem := make(chan struct{}, 64)
...
select {
case sem <- struct{}{}:
    go func() { defer func() { <-sem }(); a.handle(ctx, conn) }()
default:
    _ = conn.Close()
}
```

Location: `internal/agent/server.go:122-131`, `server.go:173-201`.

## Severity
**High** for local availability of the secret-injection path.

Galvill:
## Summary
Closes #114. Two related DoS fixes for the per-user agent's accept loop.

### 1. Deadline-before-auth
`handle` set the 10s per-conn deadline AFTER `checkPeerUid`, so the rejected-peer branch wrote `unauthorized` to a deadline-less conn. A same-user process that connects (passing the 0600/SDDL gate) and refuses to read can pin the handler goroutine indefinitely.

Moved `SetDeadline` to the first line of `handle`.

### 2. Concurrent-connection cap
`Serve` spawned an unbounded `go handle(...)` per accepted conn. A misbehaving same-user process could open thousands of half-finished connections, exhausting the agent's goroutine count and fd table.

Added a semaphore-bounded spawn keyed off `maxConcurrentAgentConns` (default 64, overridable in tests). Excess accepts are closed promptly so legitimate clients see a fast EOF and retry rather than queue forever.

## Regression test
`TestAgentRejectsExcessConnections` (cap=4) holds the cap, then verifies the (cap+1)th connection is closed at accept time (`io.EOF` on read).

## Test plan
- [x] Regression test fails on main (`i/o timeout`), passes on this branch (`io.EOF`).
- [x] `go test ./...` clean.
- [x] `GOOS=windows go vet ./...` clean.
- [ ] CI passes.

### DIFF
--- a/internal/agent/conn_limit_test.go
+++ b/internal/agent/conn_limit_test.go
@@ -1,0 +1,77 @@
+//go:build !windows
+
+package agent
+
+import (
+	"context"
+	"io"
+	"net"
+	"testing"
+	"time"
+)
+
+// TestAgentRejectsExcessConnections is the regression for security #114:
+// without a concurrent-connection cap, a same-user process can hold an
+// unbounded number of half-finished connections (each occupying a
+// goroutine and an fd in the agent), exhausting the agent's resources.
+// With the cap, connections above the limit must be closed promptly at
+// accept time.
+func TestAgentRejectsExcessConnections(t *testing.T) {
+	// Drop the cap from its default 64 to a tight value so the test is
+	// fast and doesn't depend on hitting RLIMIT_NOFILE.
+	old := maxConcurrentAgentConns
+	maxConcurrentAgentConns = 4
+	t.Cleanup(func() { maxConcurrentAgentConns = old })
+
+	a, p := newTestAgent(t, nil)
+	go a.Serve(context.Background()) //nolint:errcheck
+
+	// Wait for listener readiness via a normal Status round-trip.
+	cli := NewClient(p.Socket)
+	deadline := time.Now().Add(2 * time.Second)
+	for time.Now().Before(deadline) {
+		if _, err := cli.Status(context.Background()); err == nil {
+			break
+		}
+		time.Sleep(20 * time.Millisecond)
+	}
+
+	// Open exactly the cap's worth of connections and intentionally
+	// don't send a request — each blocks in the agent's ReadMessage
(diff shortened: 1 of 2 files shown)


Galvill:

PR #139, opened 2026-05-14
fix(sources/vault): warn loudly when tls_skip_verify is enabled (#113)

### LINKED ISSUE #113: security: Vault source 'tls_skip_verify' disables certificate validation with no warning — HTTPS_PROXY MITM
## Summary
`internal/sources/vault/vault.go:185-192` honours a `tls_skip_verify` param by setting `InsecureSkipVerify: true` on the `*http.Transport`'s `TLSClientConfig`. The transport's `Proxy` is `http.ProxyFromEnvironment`, so it consults `HTTPS_PROXY` / `HTTP_PROXY` / `NO_PROXY` from the agent's environment.

The user-facing schema help text is only `"Bool; for dev-mode against self-signed certs."` — no security warning. `asBool` accepts `"true"`, `"1"`, `"yes"` (broader than a strict bool, easier to set accidentally). There is no runtime log line when this path is taken.

## Impact
With `tls_skip_verify=true` **and** an attacker who can set `HTTPS_PROXY` in the agent's environment (i.e. set it in the shell that runs `jitenv unlock`), the agent's Vault traffic is fully visible to and modifiable by the proxy operator:
- All `X-Vault-Token` headers (AppRole login responses or static tokens) — captured.
- All KV secret values returned — captured.
- The agent fetches secrets on every mapped command, so the leak is continuous and long-lived.

The `//nolint:gosec` suppresses static-analysis warnings, removing the last automated defence.

## Fix
- Refuse to set `InsecureSkipVerify: true` unless a separate explicit boolean is set in addition to `tls_skip_verify` (e.g. `tls_skip_verify_ack_insecure = true`), or
- Always log a `WARN`-level line at client-creation time when `tls_skip_verify` is enabled: `vault source %q: TLS verification DISABLED — do not use in production`.
- Add a `tls_ca_cert` path param so the legitimate "self-signed cert" use case is served without disabling all verification.
- Update the schema `Help` text to say `INSECURE: disables certificate verification; HTTPS_PROXY can MITM all Vault traffic`.

Location: `internal/sources/vault/vault.go:185-192`.

## Severity
**High** when the param is set.

Galvill:
## Summary
Closes #113.

The vault source quietly accepted `tls_skip_verify=true` and switched to an `InsecureSkipVerify` TLS config with only a `//nolint:gosec` marker. Combined with the transport's `Proxy: http.ProxyFromEnvironment`, this let an attacker who can set `HTTPS_PROXY` in the agent's environment MITM every Vault token login and KV fetch.

## Fix
- Log a structured `WARN` line at `New()` time when `tls_skip_verify=true`. One line per source (not per fetch) so the log stays readable.
- Rewrite the schema help text from `"Bool; for dev-mode"` to an explicit INSECURE notice mentioning the `HTTPS_PROXY` MitM risk.

## Regression test
`TestNew_TLSSkipVerify_LogsWarning` captures slog output, asserts the WARN line fires for `tls_skip_verify=true`, and asserts no warning is emitted for a clean source.

## Test plan
- [x] Test fails on main, passes on this branch.
- [x] `go test ./...` clean.
- [ ] CI passes.

### DIFF
--- a/internal/sources/vault/vault.go
+++ b/internal/sources/vault/vault.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"strings"
 
@@ -74,7 +75,7 @@ func schema() []source.ParamField {
 			Enum: []string{kvV1, kvV2},
 			Help: "v1 or v2 (default v2)"},
 		{Key: "tls_skip_verify", Label: "TLS skip verify",
-			Help: "Bool; for dev-mode against self-signed certs."},
+			Help: "INSECURE: disables certificate verification. Only for dev-mode self-signed targets. Setting this in production allows HTTPS_PROXY to MITM every Vault fetch and capture tokens + secret values."},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


Galvill:

PR #140, opened 2026-05-14
fix(tui): encrypt-by-default for source params on save (#112)

### LINKED ISSUE #112: security: TUI saves source params in plaintext for any source type whose schema doesn't mark them Sensitive
## Summary
`internal/tui/save.go:91-121` (`encryptForSave`) only encrypts source params whose key appears in the type's registered schema **and** is flagged `Sensitive: true`:

```go
schema := schemaFor(sc.Type)
sensitive := map[string]struct{}{}
for _, f := range schema {
    if f.Sensitive { sensitive[f.Key] = struct{}{} }
}
for k, v := range sc.Params {
    ...
    if _, isSensitive := sensitive[k]; !isSensitive {
        continue  // SKIPPED: written to disk as plaintext
    }
    ...
}
```

Any param key that is either:
- in a source type with no registered schema (e.g. `noop` calls `sources.Register` but not `RegisterSchema`), or
- in a known-schema source but not flagged `Sensitive: true`

is written to `config.toml` **in plaintext** even if the user entered credentials there.

The TUI offers a generic key/value editor — there is no UI-level guard against the user typing a secret into a param the schema doesn't anticipate.

## Impact
- `noop` source: any param entered via the generic editor lands plaintext on disk.
- Future source authors who forget to call `RegisterSchema`, or omit `Sensitive: true` on a credential-bearing field, leak secrets silently. There is no "encrypt by default" fallback; the schema is the only gate.
- This is a footgun, not a bug: the design relies on per-source author discipline, with no runtime warning when an unrecognised string-typed param is being persisted unencrypted.

## Fix
Encrypt-by-default for **all** string params, including those without a schema entry:

```go
for k, v := range sc.Params {
    s, ok := v.(string)
    if !ok || s == "" || crypto.IsEnvelope(s) {
        continue
    }
    // Encrypt every string param. Source authors can opt specific
    // fields out via a schema flag if they're truly public (e.g. region).
    env, err := crypto.EncryptField(key, s)
    ...
}
```

Or invert the flag (`Public bool` instead of `Sensitive bool`) so the safe default is "encrypted" and authors must opt out per field. Document the change in `pkg/source/source.go`.

Location: `internal/tui/save.go:91-121`, `pkg/source/source.go` (`ParamField` definition).

## Severity
**High** — silent plaintext storage of user-entered credentials when a source author forgets a single flag.

Galvill:
## Summary
Closes #112.

`encryptForSave` previously only encrypted params whose key was both:
- in the source type's registered schema, AND
- flagged `Sensitive: true` on that schema entry.

That double-gate silently leaked any string typed into the TUI's generic key/value editor against (a) a source type with no registered schema (e.g. \`noop\`), (b) a known-schema source where the author forgot the \`Sensitive\` flag, or (c) ad-hoc params a user added through the free-form editor.

## Fix
Switch to encrypt-by-default: every non-envelope string param is encrypted on save, full stop. \`ParamField.Sensitive\` stays meaningful for the TUI (mask-while-editing + secret-key picker) but no longer governs disk encryption. The doc comment on \`ParamField\` is updated so future source authors can't be confused.

## Regression tests
- \`TestEncryptForSave_RoundTrip\` now asserts the AWS \`region\` field (previously plaintext) is encrypted on disk.
- \`TestEncryptForSave_EncryptsParamsWithoutSchema\` covers the noop-source case directly.

## Test plan
- [x] Both tests fail on \`main\`, pass on this branch.
- [ ] CI passes.

### DIFF
--- a/internal/tui/save.go
+++ b/internal/tui/save.go
@@ -7,8 +7,6 @@ import (
 
 	"github.com/gv/jitenv/internal/config"
 	"github.com/gv/jitenv/internal/crypto"
-	"github.com/gv/jitenv/internal/sources"
-	"github.com/gv/jitenv/pkg/source"
 )
 
 // saveCmd takes a snapshot of the in-memory config, re-encrypts every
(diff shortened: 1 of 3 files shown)


Galvill:

PR #141, opened 2026-05-14
fix(config,crypto): reject KDF parameters below documented floors (#111)

### LINKED ISSUE #111: security: KDF parameters in [_meta] are unauthenticated — Argon2id parameter downgrade is silent
## Summary
`internal/config/config.go` stores `Argon_Time`, `Argon_Memory_KiB`, `Argon_Threads`, and `Salt` in the `[_meta]` section of `config.toml` as plaintext, unauthenticated TOML fields. `DeriveKeyFromMeta` (`internal/config/edit.go`) reads them and applies `nz`/`nzU8` only to substitute defaults for zero values — any positive value is accepted unconditionally.

Only the `Verify` sentinel is encrypted with the derived key, and it is itself an envelope produced under attacker-controlled KDF parameters (see #X for the AAD issue: even with AAD bound, an attacker with config-write rights can recompute the sentinel locally).

## Impact
An attacker with write access to `config.toml` but no knowledge of the passphrase can:
1. Lower the cost parameters (e.g. `argon_time = 1`, `argon_memory_kib = 8`) so that the next legitimate `jitenv unlock` derives the KEK at near-zero cost.
2. Pre-compute candidate KEKs offline against a leaked memory/swap/core-dump snapshot at 100×–1000× less effort than the user-configured cost.
3. Verify candidates by trying to decrypt one of the legitimate envelopes elsewhere in the config (the user's actual data) — the attacker does **not** need to forge a sentinel.

Result: a credible offline brute-force path against weak passphrases that the user explicitly tried to harden via Argon2id parameters.

## Fix
- In `DeriveKeyFromMeta`, refuse to derive a key from `Meta` whose parameters are below the project's documented minimums (e.g. `ArgonTime < 2`, `ArgonMemoryKiB < 19_456` = OWASP minimum 19 MiB, `ArgonThreads < 1`, `SaltLen < 16`). Return a clear error and instruct the user to re-init.
- Bind the meta fields into the AAD of the `Verify` envelope (and ideally every envelope) so any parameter tampering invalidates the sentinel. This is symbiotic with the AAD fix above.

Location: `internal/config/config.go:24-31`, `internal/config/edit.go:65-70`, `internal/crypto/kdf.go`.

## Severity
**High** in the "attacker has config-write" threat model that the rest of the system implicitly assumes (config is mode 0600 but backups, NFS mounts, malicious installers, etc. all defeat that).

Galvill:
## Summary
Closes #111.

`[_meta]` holds `argon_time` / `argon_memory_kib` / `argon_threads` / `salt` as plaintext, unauthenticated TOML fields. A config-write attacker could drop the parameters to near-zero so the next legitimate unlock derives the KEK cheaply — and any offline brute-force against a leaked memory/swap/core-dump becomes correspondingly cheap.

## Fix
Add minimum constants in `internal/crypto/kdf.go` and a guard in `DeriveKeyFromMeta`:

| param | min |
|---|---|
| `MinArgonTime` | 2 |
| `MinArgonMemKiB` | 19 × 1024 (OWASP) |
| `MinArgonThreads` | 1 |
| `MinSaltLen` | 16 |

Reject configs below the floors at derive time with a clear error that points the user at `jitenv config init`. Default-init params already exceed all minimums, so existing configs continue to work unchanged.

## Regression test
`TestDeriveKeyFromMeta_RejectsWeakArgonParams` mutates a freshly-init'd config's `[_meta]` to set `argon_time=1` / `argon_memory_kib=8`, asserts `DeriveKeyFromMeta` returns an error mentioning the offending param.

## Test plan
- [x] Test fails on `main` (incorrect-passphrase generic error), passes on this branch.
- [x] `go test ./...` clean.
- [ ] CI passes.

### DIFF
--- a/internal/config/edit.go
+++ b/internal/config/edit.go
@@ -67,6 +67,23 @@ func DeriveKeyFromMeta(c *Config, passphrase []byte) ([]byte, error) {
 		MemKiB:  nz(c.Meta.ArgonMemoryKiB, crypto.DefaultArgonMemKiB),
 		Threads: nzU8(c.Meta.ArgonThreads, crypto.DefaultArgonThreads),
 	}
+	// Reject KDF params below documented floors. Without this check a
+	// config-write attacker can drop argon_time to 1 and argon_memory_kib
+	// to a few KiB so the next derive (and any offline brute-force
+	// against a leaked memory/swap dump) costs almost nothing
+	// (security #111).
+	if params.Time < crypto.MinArgonTime {
+		return nil, fmt.Errorf("config _meta.argon_time=%d is below minimum %d; re-init with `jitenv config init`", params.Time, crypto.MinArgonTime)
+	}
+	if params.MemKiB < crypto.MinArgonMemKiB {
+		return nil, fmt.Errorf("config _meta.argon_memory_kib=%d is below minimum %d (OWASP floor); re-init with `jitenv config init`", params.MemKiB, crypto.MinArgonMemKiB)
+	}
+	if params.Threads < crypto.MinArgonThreads {
+		return nil, fmt.Errorf("config _meta.argon_threads=%d is below minimum %d; re-init with `jitenv config init`", params.Threads, crypto.MinArgonThreads)
+	}
+	if len(salt) < crypto.MinSaltLen {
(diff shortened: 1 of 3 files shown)


Galvill:

PR #142, opened 2026-05-14
fix(crypto,config,tui,seed): bind envelopes to per-slot AAD (#110)

### LINKED ISSUE #110: security: AEAD envelopes have no associated data — ciphertexts can be transplanted between fields
## Summary
`internal/crypto/aead.go` calls `aead.Seal(nil, nonce, plaintext, nil)` and `aead.Open(nil, nonce, ct, nil)` — both pass `nil` as the AEAD additional-data parameter. As a result, every `enc:v1:…` envelope on disk is decryptable in any context, with no binding to the field/source/secret it belongs to.

## Impact
An attacker who can write to `config.toml` (but does not know the passphrase) can move a ciphertext from one slot to another, e.g.:
- Copy a `vault_token = "enc:v1:<blobA>"` envelope into `aws_secret_access_key = "enc:v1:<blobA>"`.
- Swap two source params, two secrets in different bags, or move a secret value into a source param slot.

`DecryptStringsInPlace` walks every string in the config and decrypts each envelope in place, with no knowledge of where the envelope lives. The agent then hands the wrong plaintext to the wrong consumer:
- An AWS source receiving a Vault token as `aws_secret_access_key` will sign requests against AWS with attacker-chosen bytes — useful if the attacker wants to drive a specific error/log.
- A Vault source receiving an AWS key as its token will send that key to a Vault endpoint in an `X-Vault-Token` header — direct exfil of the AWS credential to whatever Vault address is configured (which is plaintext in the config and modifiable by the same attacker).

This crosses the boundary from passive-disclosure (which the passphrase protects) into active-misuse (which it currently does not).

## Reproducer
1. `jitenv config init`, add two sources: `[sources.vault.params] token = "..."` and `[sources.aws.params] secret_access_key = "..."`. Save.
2. Edit `config.toml`: copy the base64 envelope from `vault.params.token` into `aws.params.secret_access_key`. Save.
3. Run a mapped command that uses the `aws` source. The agent decrypts the Vault token, hands it to the AWS source as a secret key, and the AWS SDK signs requests with the wrong material. Swap directions to send AWS material to a Vault endpoint.

## Fix
Add an associated-data argument to `Seal`/`Open` and bind every envelope to its semantic location:
- Source params: `AD = "source/" + sourceName + "/" + paramKey`
- Secrets:       `AD = "secret/" + bag + "/" + key`
- Meta verify:   `AD = "meta/verify"`

`DecryptStringsInPlace` already knows the path it's at (it can be threaded the context). Existing envelopes need a one-shot migration: on read, accept envelopes with empty AD for backward compat once, then re-seal with the correct AD on next save.

Location: `internal/crypto/aead.go:25,45`; threaded through `internal/crypto/envelope.go` and `internal/config/decrypt.go`.

## Severity
**High** — requires config-write, but the AEAD's tag is doing less work than the threat model implies.

Galvill:
## Summary
Closes #110.

XChaCha20-Poly1305 was sealing with `nil` associated data, so every `enc:v1:` blob on disk was decryptable in any slot. An attacker with config-write rights could move a \`vault_token\` envelope into an \`aws_secret_access_key\` slot and the agent would hand the wrong plaintext to the wrong consumer.

## Fix
Introduce \`enc:v2:\` envelopes bound via the AEAD tag to a canonical AAD string derived from the value's location:

| slot | AAD |
|---|---|
| source param | \`src.<sourceName>.<paramKey>\` |
| local secret | \`secret.<bag>.<key>\` |
| meta verify | \`meta.verify\` |

### API changes (internal module only)
- \`crypto.Seal\` / \`crypto.Open\` now take an \`ad []byte\` argument.
- \`crypto.EncryptField\` / \`crypto.DecryptField\` now take an \`ad string\`.
- \`crypto.DecryptStringsInPlace\` now takes a \`ctx string\` prefix that the walker concatenates with the map key.
- \`crypto.AAD(parts...)\` builds canonical dotted strings; both sides must use it (or the helpers in \`config\`) to stay in sync.

### Backward compatibility
Reads still accept \`enc:v1:\` envelopes (legacy, no AAD), so existing on-disk configs continue to load. The next save rewrites them as \`enc:v2:\` with proper binding.

## Regression tests
- \`crypto.TestEnvelopeRejectsTransplant\` — \`enc:v2\` refuses to decrypt under any AAD other than the seal-time one.
- \`crypto.TestEnvelopeAcceptsLegacyV1\` — \`enc:v1\` still decrypts (legacy compat).
- \`config.TestDecryptInPlace_RejectsTransplantedEnvelope\` — end-to-end: a token-bound envelope dropped into a different source slot is rejected by \`DecryptInPlace\`.

## Test plan
- [x] All transplant tests fail on main, pass on this branch.
- [ ] CI passes (incl. all e2e seed permutations).

### DIFF
--- a/e2e/seed/seed.go
+++ b/e2e/seed/seed.go
@@ -196,11 +196,11 @@ func run(o runOpts) error {
 // expands the bag. fooVal/barVal let the caller produce distinguishable
 // fixtures (used by local-alt to differentiate before/after a reload).
 func applyLocal(cfg *config.Config, key []byte, scriptPath, fooVal, barVal string) error {
-	foo, err := crypto.EncryptField(key, fooVal)
+	foo, err := crypto.EncryptField(key, fooVal, config.SecretAAD("demo", "FOO"))
 	if err != nil {
 		return err
 	}
-	bar, err := crypto.EncryptField(key, barVal)
+	bar, err := crypto.EncryptField(key, barVal, config.SecretAAD("demo", "BAR"))
 	if err != nil {
 		return err
 	}
@@ -226,15 +226,15 @@ func applyLocal(cfg *config.Config, key []byte, scriptPath, fooVal, barVal strin
 // (instead of an exact Path) and a bag with three keys that all expand
 // via a single empty-Name VarRef.
 func applyLocalGlob(cfg *config.Config, key []byte, globPath string) error {
-	foo, err := crypto.EncryptField(key, "value-from-local-foo")
+	foo, err := crypto.EncryptField(key, "value-from-local-foo", config.SecretAAD("demo", "FOO"))
 	if err != nil {
 		return err
 	}
-	bar, err := crypto.EncryptField(key, "value-from-local-bar")
+	bar, err := crypto.EncryptField(key, "value-from-local-bar", config.SecretAAD("demo", "BAR"))
 	if err != nil {
 		return err
 	}
-	baz, err := crypto.EncryptField(key, "value-from-local-baz")
+	baz, err := crypto.EncryptField(key, "value-from-local-baz", config.SecretAAD("demo", "BAZ"))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


Galvill:

PR #144, opened 2026-05-14
fix(agent,shell): verify runtime dir ownership + 0700 mode (#117)

### LINKED ISSUE #117: security: /tmp/jitenv-<uid> fallback directory ownership not verified — pre-creation enables DoS
## Summary
On Linux/macOS when `\$XDG_RUNTIME_DIR` is unset, `internal/agent/paths_unix.go:19` falls back to `/tmp/jitenv-<uid>/` and `DefaultPaths()` (`internal/agent/paths.go:61`) calls `os.MkdirAll(p.Dir, 0700)`.

`os.MkdirAll` succeeds silently when the directory already exists, **regardless of who owns it or its current mode**. An attacker running as a different user can pre-create `/tmp/jitenv-1000/` before the victim (uid 1000) first runs jitenv. The attacker now owns the directory.

## Impact
Once the attacker owns the directory containing `agent.sock`, `agent.pid`, and `agent.log`, they can:
- `unlink` `agent.sock` to break IPC (denial of service: legitimate clients see `connect: no such file or directory`).
- Replace `agent.sock` with a symlink or a different socket. The agent's own `SO_PEERCRED` check rejects the wrong UID, so the attacker cannot impersonate the agent successfully — but the victim's clients still see connection errors.
- Truncate `agent.log` to suppress diagnostics.

This is also a TOCTOU root cause for #X (chmod-after-bind window) — even if the socket creation race were fixed, the directory itself is already attacker-owned, so the per-file mode is moot.

## Reproducer
```sh
# As uid 2000, before victim (uid 1000) runs jitenv:
mkdir /tmp/jitenv-1000
# After victim's agent creates the socket, attacker can:
rm /tmp/jitenv-1000/agent.sock   # works — attacker owns dir
```

## Fix
After `MkdirAll`, `os.Lstat` the resulting directory and verify ownership + mode before continuing:

```go
st, err := os.Lstat(dir)
if err != nil { return err }
sys, ok := st.Sys().(*syscall.Stat_t)
if !ok || sys.Uid != uint32(os.Getuid()) {
    return fmt.Errorf("runtime dir %s not owned by current user", dir)
}
if st.Mode().Perm() != 0o700 {
    return fmt.Errorf("runtime dir %s has mode %o, want 0700", dir, st.Mode().Perm())
}
```

This is the pattern used by `gpg-agent`, `ssh-agent`, and OpenSSH for their runtime sockets.

Location: `internal/agent/paths.go:55-65`, `internal/agent/paths_unix.go:15-22`.

## Severity
**Medium** — DoS only, no secret exfil, but trivial to trigger on shared systems where `/tmp` is the fallback.

Galvill:
## Summary
Closes #117.

\`agent.DefaultPaths\` called \`os.MkdirAll(0700)\` on the runtime dir but took no further action. MkdirAll silently accepts a pre-existing directory regardless of who owns it or what mode it has. On the \`/tmp/jitenv-<uid>\` fallback (XDG_RUNTIME_DIR unset) an attacker with write access to /tmp can pre-create the directory at any mode and later unlink the agent's socket, replace \`agent.log\`, or plant a hostile pid file.

## Fix
- Add \`verifyRuntimeDir\` in the agent package. Unix variant: Lstat, require uid==self + mode==0700. Windows variant: no-op (LOCALAPPDATA is per-user, named-pipe ACL is the load-bearing gate).
- Call it from \`DefaultPaths\` after MkdirAll. Errors propagate so startup refuses rather than binding a 0600 socket inside an attacker-owned directory.

The bash and zsh hook snippets pre-create the wrapper subtree with plain \`mkdir -p\`, which inherits the user's umask (typically 022 → 0755) and would trip the new check. Wrap each mkdir in a subshell with \`umask 077\` so every intermediate lands at 0700.

## Regression tests
- \`TestDefaultPaths_RejectsWrongModeRuntimeDir\` — pre-existing 0755 dir is refused.
- \`TestDefaultPaths_AcceptsCorrectlyOwnedDir\` — fresh 0700 dir is accepted.

## Test plan
- [x] \`GOOS=windows go vet ./...\` clean.
- [ ] CI passes.

### DIFF
--- a/internal/agent/paths.go
+++ b/internal/agent/paths.go
@@ -56,9 +56,21 @@ func ResolvePaths() Paths {
 // dir (0700) so callers that need to bind a socket or open a log
 // file can rely on it existing. Use this from agent startup, chpwd,
 // etc.
+//
+// After MkdirAll, verify the resulting directory is owned by the
+// current user with mode 0700 (security #117). MkdirAll silently
+// accepts a pre-existing directory regardless of ownership/mode, so
+// without this check an attacker on a shared host (think the
+// /tmp/jitenv-<uid> fallback) could pre-create the dir and later
+// unlink the agent's socket or replace the log. The check is
+// platform-split; the Windows variant is a no-op because Windows
+// ACLs differ and the LOCALAPPDATA path is already per-user.
 func DefaultPaths() (Paths, error) {
 	p := ResolvePaths()
 	if err := os.MkdirAll(p.Dir, 0700); err != nil {
+		return Paths{}, err
+	}
+	if err := verifyRuntimeDir(p.Dir); err != nil {
 		return Paths{}, err
 	}
 	return p, nil
--- a/internal/agent/paths_unix.go
+++ b/internal/agent/paths_unix.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"syscall"
 )
 
 // runtimeBaseDir returns the per-user runtime directory for the agent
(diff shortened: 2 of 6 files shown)


Galvill:

PR #145, opened 2026-05-14
fix(agent/socket_unix): set umask before bind so socket lands at 0600 atomically (#118)

### LINKED ISSUE #118: security: Unix socket created with chmod-after-bind window — defence-in-depth gap before peer check
## Summary
`internal/agent/socket_unix.go:19-27` binds the listener with `net.ListenUnix` (which uses the process umask) and then calls `os.Chmod(path, 0600)`. Between the bind and the chmod, the socket file exists with `(0666 & ~umask)` permissions — typically `0666 & ~022 = 0644` (world-readable, no connect right) or, if the umask is unusual, `0666` (world-connectable).

The `checkPeerUid` (`SO_PEERCRED` on Linux, `LOCAL_PEERCRED` on Darwin) is the actual access control, so a wrong-UID connect during the window is still rejected at the agent layer. But the comment in `socket_unix.go:12-13` describes the 0600 mode as the **first** line of defence — which is temporarily false.

## Impact
- The window is microseconds, so practical exploitability is low.
- It becomes load-bearing if `checkPeerUid` ever regresses (e.g. a refactor moves the check below `WriteMessage`, as already noted in the unauthorised-write issue).
- It is detectable: a same-host attacker spinning on `connect(2)` against the path during agent start can land a connection without the 0600 gate, falling back to peer-cred alone.

## Fix
Set the umask before `net.ListenUnix` and restore it immediately after:

```go
func listenSocket(path string) (*net.UnixListener, error) {
    old := syscall.Umask(0o177)
    defer syscall.Umask(old)
    l, err := net.ListenUnix("unix", &net.UnixAddr{Name: path, Net: "unix"})
    ...
}
```

Umask is process-wide, so this is safe only if `Listen` is called before other goroutines do file I/O. `Listen()` runs before `Serve()` so this holds. Alternatively, create the socket file at a `.tmp` path with `O_CREAT|O_EXCL|0600` first, then bind+rename — but the umask trick is simpler.

Location: `internal/agent/socket_unix.go:14-30`.

## Severity
**Medium** — defence-in-depth only; not exploitable on its own.

Galvill:
## Summary
Closes #118.

\`listenSocket\` used \`net.ListenUnix\` followed by \`os.Chmod(0600)\`. For the brief window between bind and chmod the socket file existed at \`(0666 & ~umask)\` — typically 0644 with the default 0022 umask, or as open as 0666 with an unusual umask=0. \`checkPeerUid\` is the load-bearing gate so a wrong-uid client could never get past the agent, but the defence-in-depth mode bits weren't actually being applied at bind time.

## Fix
Wrap the \`ListenUnix\` call in a \`syscall.Umask(0o177)\` / restore pair so the kernel applies 0600 directly. The follow-up \`os.Chmod\` stays as belt-and-suspenders for any platform where AF_UNIX bind ignores umask (none observed).

## Regression test
\`TestListenSocketMode_AtomicallyZeroOther\` forces a permissive umask (0) before \`listenSocket\` and asserts the resulting file is still 0600 — only true when the function sets its own restrictive umask before the bind.

## Test plan
- [ ] CI passes.

### DIFF
--- a/internal/agent/socket_unix.go
+++ b/internal/agent/socket_unix.go
@@ -6,17 +6,32 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"syscall"
 )
 
-// listenSocket binds the per-user Unix socket at path and chmods it
-// 0600. The 0600 mode is load-bearing: peer-credential checks happen
-// in checkPeerUid, but the filesystem permission is the first line of
-// defence against an unrelated uid even attempting to connect.
+// listenSocket binds the per-user Unix socket at path with mode 0600
+// applied atomically via the process umask. The 0600 mode is load-
+// bearing: peer-credential checks happen in checkPeerUid, but the
+// filesystem permission is the first line of defence against an
+// unrelated uid even attempting to connect.
 //
-// Returns a net.Listener so callers stay platform-agnostic; the Windows
-// counterpart in socket_windows.go uses a named-pipe listener instead.
+// Setting the umask before bind closes the chmod-after-bind window
+// (security #118): with the previous order, the socket existed for a
+// few microseconds at (0666 & ~umask, typically 0644) before Chmod
+// tightened it. A same-host attacker spinning on connect(2) could
+// land a connection in that window. checkPeerUid is the actual access
+// gate so this never let a wrong-uid client past the agent, but the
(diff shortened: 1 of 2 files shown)


Galvill:

PR #146, opened 2026-05-14
fix(sources/vault): reject path-traversal refs at Fetch time (#119)

### LINKED ISSUE #119: security: Vault readPath concatenates user-controlled ref into HTTP path — '..' traversal can escape configured mount
## Summary
`internal/sources/vault/vault.go:303-317` builds the Vault HTTP path by string concatenation:

```go
return v.mount + "/data/" + name   // KV v2
return v.mount + "/" + name        // KV v1
```

`name` comes from `VarRef.Ref` in the user's config, after a one-shot `TrimPrefix(name, mountPrefix)`. `..` components are not stripped. Vault's HTTP router normalises `..` in the path before dispatching.

## Impact
A `VarRef.Ref` such as `"../../sys/policies/acl/default"` produces a URL of `<mount>/data/../../sys/policies/acl/default` → after Vault path normalisation → `sys/policies/acl/default`. The configured mount restriction is bypassed; the request hits any Vault API the token has policy access to (audit logs, policies, identity endpoints, even `sys/raw` if the token is sufficiently privileged).

This is not direct secret exfil — jitenv only consumes JSON keys named in `Lookup` — but it widens the blast radius of a misconfigured token from "this mount" to "anything the policy permits". It also breaks the user's mental model: they think the mount field bounds what jitenv can read.

## Reproducer
```toml
[sources.vault.params]
addr  = "https://vault.example/"
token = "..."
mount = "secret"

[[mappings]]
match = "/usr/local/bin/foo"
[[mappings.vars]]
name = "X"
source = "vault"
ref = "../../sys/policies/acl/default"
key = "rules"
```

Running the mapped command issues `GET <addr>/v1/secret/data/../../sys/policies/acl/default` → Vault serves `/v1/sys/policies/acl/default`.

## Fix
Reject path traversal in `readPath` (or in `Validate`):

```go
if name == "" || strings.Contains(name, "..") || strings.HasPrefix(name, "/") {
    return nil, fmt.Errorf("vault: invalid ref %q", name)
}
```

Or `path.Clean` and verify the result has no leading `..`. Prefer rejection — silent normalisation hides the mistake.

Location: `internal/sources/vault/vault.go:303-317` and `Validate` for `VarRef`.

## Severity
**Medium** — requires the Vault token's policy to cover the target path, and requires config write or social engineering of the user, but worth fixing.

Galvill:
## Summary
Closes #119.

\`readPath\` concatenates the configured mount with the user's \`VarRef.Ref\`, doing only a one-shot \`TrimPrefix\` of the mount segment. A ref like \`"../../sys/policies/acl/default"\` produced an HTTP path that Vault's server-side router normalised back into \`sys/…\`, silently escaping the configured mount and reaching any path the token's policy permitted.

## Fix
Validate \`ref.ID\` at the top of \`Fetch\` and refuse any input whose slash-separated segments include \`..\`. This catches the obvious \`"../../foo"\` and \`"foo/../bar"\` shapes plus the standalone \`".."\`, and runs before the network call so the rejection is fast and obviously intentional. Clean refs are unaffected.

## Regression test
\`TestFetch_RejectsPathTraversal\` uses a pre-cancelled context: a clean ref dies with \`context.Canceled\` (network was attempted), a traversal ref dies with the new \`"contains path traversal"\` error (rejected at validation).

## Test plan
- [ ] CI passes.

### DIFF
--- a/internal/sources/vault/vault.go
+++ b/internal/sources/vault/vault.go
@@ -260,6 +260,13 @@ func (v *vaultSource) Fetch(ctx context.Context, ref source.SecretRef) (map[stri
 	if ref.ID == "" {
 		return nil, fmt.Errorf("vault: ref.ID (KV path) is required")
 	}
+	// Reject path-traversal refs before hitting the network (security
+	// #119). Vault's HTTP router normalises `..` server-side, so a ref
+	// like "../../sys/policies" would otherwise escape the configured
+	// mount and reach any path the token's policy permits.
+	if hasPathTraversal(ref.ID) {
+		return nil, fmt.Errorf("vault: ref %q contains path traversal segment", ref.ID)
+	}
 	cli, err := v.newClient(ctx)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


Galvill:

PR #148, opened 2026-05-14
fix(crypto/passphrase): zero buffers + constant-time compare (#126)

### LINKED ISSUE #126: security: PromptPassphrase leaks passphrase bytes — pw2 never zeroed, comparison via string() copies, no constant-time compare
## Summary
`internal/crypto/passphrase.go:34-44`:

```go
pw, err := term.ReadPassword(int(syscall.Stdin))
...
if confirm {
    pw2, err := term.ReadPassword(int(syscall.Stdin))
    ...
    if string(pw) != string(pw2) {
        return nil, errors.New("passphrases do not match")
    }
}
return pw, nil
```

Three issues:
1. `pw2` is allocated by `term.ReadPassword` and is never zeroed. It lingers in the GC heap on both the match and mismatch paths.
2. `pw` is not zeroed on the mismatch error return inside `PromptPassphrase`. The caller's `defer zeroBytes(pw)` is irrelevant because the returned value is `nil` — the local `pw` slice goes out of scope without being cleared.
3. `string(pw) != string(pw2)` creates two **immutable** Go string copies. Strings cannot be zeroed; their backing bytes live until GC overwrites them. The comparison is also non-constant-time (irrelevant for local passphrases, but cheap to fix).

## Impact
After `jitenv config init` with a mismatch (or even a successful match), the passphrase bytes remain recoverable from the process heap for an indeterminate time. Combined with core dumps, swap, or `/proc/<pid>/mem`, this gives an attacker a longer window than the explicit `defer zeroBytes(key)` discipline elsewhere implies.

## Fix
```go
func PromptPassphrase(prompt string, confirm bool) ([]byte, error) {
    ...
    pw, err := term.ReadPassword(int(syscall.Stdin))
    if err != nil { return nil, err }
    if confirm {
        ...
        pw2, err := term.ReadPassword(int(syscall.Stdin))
        if err != nil { zeroBytes(pw); return nil, err }
        defer zeroBytes(pw2)
        if subtle.ConstantTimeCompare(pw, pw2) != 1 {
            zeroBytes(pw)
            return nil, errors.New("passphrases do not match")
        }
    }
    return pw, nil
}
```

`subtle.ConstantTimeCompare` operates on byte slices — no string conversion, no immutable copies, constant-time as a bonus.

Location: `internal/crypto/passphrase.go:30-50`.

## Severity
**Medium** — memory-hygiene gap in the most sensitive code path.

Galvill:
## Summary
Closes #126.

\`PromptPassphrase\` abandoned the confirmation buffer (\`pw2\`) without zeroing, and used \`string(pw) != string(pw2)\` which spawns two immutable Go string copies of both passphrases — those copies live in the heap until next GC and cannot be cleared. The first passphrase buffer (\`pw\`) was also leaked on the mismatch error path.

## Fix
- Add a package-private \`zeroPassphrase\` helper.
- Zero \`pw\` on mismatch / second-read-error paths.
- \`defer zeroPassphrase(pw2)\` so it never outlives the function.
- Replace the string comparison with \`subtle.ConstantTimeCompare\` on the raw byte slices. No mid-function string conversion of the passphrase remains.

## Test plan
- [x] \`go test ./...\` clean (no behavioural change to the success path).
- [ ] CI passes.

### DIFF
--- a/internal/crypto/passphrase.go
+++ b/internal/crypto/passphrase.go
@@ -1,11 +1,26 @@
 package crypto
 
 import (
+	"crypto/subtle"
 	"errors"
 	"fmt"
 
 	"golang.org/x/term"
 )
+
+// zeroPassphrase overwrites the bytes of a passphrase slice in place.
+// Used everywhere PromptPassphrase abandons a buffer (mismatched
+// confirm, empty-passphrase reject, read error after first input) so
+// the cleartext lifespan in the heap is bounded. Go strings are
+// immutable and unzeroable; the comparison path therefore uses
+// subtle.ConstantTimeCompare on the raw byte slices rather than
+// `string(pw) != string(pw2)`, which would spawn two unzeroable
+// string copies (security #126).
+func zeroPassphrase(b []byte) {
+	for i := range b {
+		b[i] = 0
+	}
+}
 
 // PromptPassphrase reads a passphrase from the controlling terminal
 // without echo. The terminal is opened directly (rather than via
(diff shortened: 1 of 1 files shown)


Galvill:

PR #149, opened 2026-05-14
fix(cli/unlock): build a real resolver in --foreground mode (#131)

### LINKED ISSUE #131: security: 'jitenv unlock --foreground' starts an agent with nil resolver — silently injects no secrets
## Summary
`internal/cli/unlock.go:49-59` (foreground branch) constructs an `agent.Agent` with `nil` as the resolver and starts it. The derived key is still in scope and is `defer zeroBytes`-d correctly — but it is never used to load the config or build a resolver. Every `OpFetchEnv` returned by this agent is an empty map.

```go
if foreground {
    ag := agent.NewAgent(paths, idle, nil)  // ← nil resolver
    if err := ag.Listen(); err != nil { ... }
    return ag.Serve(ctx)
}
```

`server.go:dispatch` correctly handles `r == nil` defensively (no panic), so the agent serves connections but tells every client "no mapping" / "no secrets".

## Impact
A user (or operator) running `jitenv unlock --foreground` to debug or as a dry-run gets an agent that **looks** like it works — `jitenv status` reports it as up, `is-mapped` returns "not mapped" for everything — but injects nothing into any mapped command. Commands run silently against whatever credentials happen to be in the ambient shell, which is exactly the failure mode jitenv is designed to prevent.

This isn't an external-attacker vulnerability; it's a fail-open design defect.

## Fix
In the foreground branch, build a resolver the same way the background spawn does (mirror `internal/cli/agent_internal.go`'s `loadAndBuild` closure), or remove the `--foreground` flag if it isn't a supported usage. If kept, gate behind `--dev` and log a loud warning that no secrets will be injected.

Location: `internal/cli/unlock.go:49-59`.

## Severity
**Low** — affects users who explicitly opt in to `--foreground`.

Galvill:
Closes #131. See commit message.

### DIFF
--- a/internal/cli/unlock.go
+++ b/internal/cli/unlock.go
@@ -47,7 +47,29 @@ func newUnlockCmd() *cobra.Command {
 			idle := parseIdle(cfg.Agent.IdleTimeout)
 
 			if unlockForeground {
-				ag := agent.NewAgent(paths, idle, nil)
+				// Build a real resolver so the foreground agent
+				// actually injects secrets — previously it was started
+				// with a nil resolver and silently returned empty env
+				// for every mapped command (security #131).
+				loadAndBuild := func() (agent.Resolver, error) {
+					c, err := config.Load(cfgPath)
+					if err != nil {
+						return nil, err
+					}
+					if err := config.DecryptInPlace(c, key); err != nil {
+						return nil, err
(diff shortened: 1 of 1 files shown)


Galvill:

PR #150, opened 2026-05-14
fix(cli/agent): drop source-name list from INFO log line (#129)

### LINKED ISSUE #129: security: agent.log records config path and source-name list at INFO — operational metadata disclosure
## Summary
`internal/cli/agent_internal.go:40,72`:
```go
slog.Info("agent starting", "config", cfgArg, "idle", idle.String())
slog.Info("agent listening", "socket", paths.Socket, "sources", res.Sources())
```

`cfgArg` is the resolved config file path. `res.Sources()` is the list of configured source **names** (e.g. `"aws-prod"`, `"vault-staging"`, `"local"`). Both go to `agent.log` (0600, user-only) at INFO.

The log file mode is correct and the contents are metadata, not secret material. The disclosure is to whoever can read the log — i.e. the user themselves, plus root on Unix.

## Impact
- Source names like `aws-prod`, `vault-eu-staging`, `kube-secrets-tier1` constitute operational metadata about the user's secret infrastructure.
- Anyone who gains same-user code execution can read `agent.log` and learn the names/types of backends in use — useful for phishing, follow-on lateral movement, or scoping a larger compromise.
- This is a minor information leak; it does not unlock secrets.

## Fix
- Drop the `sources` field to DEBUG level (or omit by default). Keep the bare "agent listening" line.
- Keep the config path at INFO — it isn't sensitive on its own.
- Document the log location and that it includes source names; or provide a `--quiet` / `JITENV_LOG_LEVEL` knob.

Location: `internal/cli/agent_internal.go:40,72`.

## Severity
**Low** — defence-in-depth.

Galvill:
Closes #129. See commit.

### DIFF
--- a/internal/cli/agent_internal.go
+++ b/internal/cli/agent_internal.go
@@ -69,7 +69,15 @@ func newAgentInternalCmd() *cobra.Command {
 			ctx, cancel := agent.AwaitSignal(context.Background())
 			defer cancel()
 			defer ag.Shutdown()
-			slog.Info("agent listening", "socket", paths.Socket, "sources", res.Sources())
+			// The startup line previously logged res.Sources() at INFO,
+			// which dumps the configured source names into agent.log.
+			// Source names ("aws-prod", "vault-staging", "kube-secrets-
+			// tier1") are operational metadata that constitute a useful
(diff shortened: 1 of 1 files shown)


Galvill:

PR #151, opened 2026-05-14
fix(shell/snippets): shell-quote $resolved in BUFFER/eval rewrites (#123)

### LINKED ISSUE #123: security: bash/zsh BUFFER rewrites break for paths containing double-quote characters
## Summary
The hook snippets rebuild the user's command via:

`internal/shell/snippets/bash.sh:125`:
```bash
eval "jitenv run \"$resolved\"$rest"
```

`internal/shell/snippets/zsh.sh:91`:
```zsh
BUFFER="jitenv run \"$resolved\"$rest"
```

`$resolved` is the canonical absolute path of the mapped binary. It is interpolated into a double-quoted string with no shell-escaping. Filenames on Linux/macOS may legally contain `"` characters. A path like `/opt/my"tool/run` produces:

```
jitenv run "/opt/my"tool/run" --flag
```

The shell parses this as three tokens with broken quoting: `jitenv`, `run`, `/opt/my` plus `tool/run" --flag` — eval fails with a syntax error or, worse, executes a malformed command line. The user's command silently does not run, or runs with corrupted args.

This is not exploitable to inject code (the path is resolved by jitenv itself, not user-controlled at the shell layer in a meaningful way) — but it's a correctness gap that ought to be closed before any future change opens it up to actual injection.

## Fix
- bash: use `printf '%q' "$resolved"` to safe-quote, then pass un-quoted into eval. Or restructure to avoid `eval` entirely (pass args via `set --` and a function).
- zsh: use the `${(q+)resolved}` flag, which produces a shell-quoted form:
  ```zsh
  BUFFER="jitenv run ${(q+)resolved}$rest"
  ```

Location: `internal/shell/snippets/bash.sh:125`, `internal/shell/snippets/zsh.sh:91`.

## Severity
**Medium** — correctness defect with a small but real adjacency to injection if the path-resolution layer is ever extended to accept user-influenced fragments.

Galvill:
Closes #123. See commit.

### DIFF
--- a/internal/shell/snippets/bash.sh
+++ b/internal/shell/snippets/bash.sh
@@ -122,7 +122,13 @@ __jitenv_debug_trap() {
             __jitenv_log "branch=case0 (mapped → jitenv run)"
             local rest="${cmd#"$first_raw"}"
             __JITENV_REENTRY=1
-            eval "jitenv run \"$resolved\"$rest"
+            # printf %q produces a shell-safe quoted form of the path,
+            # so a filename containing `"`, `$`, backtick, or other
+            # bash-active characters (legal on Linux) can't break the
+            # eval string's quoting (security #123).
+            local quoted_resolved
+            printf -v quoted_resolved '%q' "$resolved"
+            eval "jitenv run $quoted_resolved$rest"
             unset __JITENV_REENTRY
             return 1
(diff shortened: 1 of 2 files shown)


Galvill:

PR #153, opened 2026-05-15
fix(agent): drain in-flight request handlers on Shutdown (#134)

### LINKED ISSUE #134: security: agent signal handler does not drain in-flight requests — slow fetches mid-shutdown surface as connection errors
## Summary
`internal/agent/server.go:116-120` (`AwaitSignal` goroutine) and `:278-289` (`OpLock` shutdown):

```go
go func() {
    time.Sleep(50 * time.Millisecond)  // grace for the OpLock reply to flush
    a.cancel()
}()
```

`a.cancel()` closes the listener and propagates context cancellation. Active `handle` goroutines inside `dispatch` (mid `FetchEnv`) honour the cancellation only for operations that check `ctx.Done()` — specifically the source `Fetch` paths via `FetchEnv`. The `WriteMessage` for the response has no guarantee of completing if the listener has already been torn down or the client has already given up.

There's no `WaitGroup` tracking active handlers, so `Shutdown` doesn't wait for them.

## Impact
- On lock or SIGTERM mid-fetch, a slow Vault/AWS round trip is aborted and the client receives a half-written response or `EOF`. From the user's perspective, a mapped command at that exact moment fails with a confusing IPC error instead of "agent is locking, try again".
- No data corruption, no secret exposure.
- Worst-case: a denial-of-service tool can repeatedly trigger `OpLock` mid-fetch to garble the user's commands. Requires same-user privilege, so not a meaningful escalation.

## Fix
- Track active handlers in a `sync.WaitGroup`. On shutdown, allow up to e.g. 2 seconds for in-flight `WriteMessage` calls to complete before closing the listener.
- Pattern is `net/http.Server.Shutdown`.

Location: `internal/agent/server.go:116-144,173-201,278-289`.

## Severity
**Low** — robustness improvement, minor user-visible papercut.

Galvill:
Closes #134. See commit.

### DIFF
--- a/internal/agent/server.go
+++ b/internal/agent/server.go
@@ -12,6 +12,10 @@ import (
 	"time"
 )
 
+// shutdownDrainTimeout caps how long Shutdown waits for in-flight
+// request handlers to complete before forcing exit. Tunable from tests.
+var shutdownDrainTimeout = 2 * time.Second
+
 // Agent is the per-user secret-fetching daemon.
 type Agent struct {
 	paths    Paths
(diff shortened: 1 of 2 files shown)


Galvill:

PR #154, opened 2026-05-15
fix(agent): use OS-level advisory lock on the pidfile (#130)

### LINKED ISSUE #130: security: pidfile-based 'already running' check is non-atomic — double-spawn race and PID reuse on Windows
## Summary
`internal/agent/server.go:89-101` and `internal/agent/daemonize_unix.go:23-25` (and Windows analogue) implement the agent-is-already-running guard as:

```
ReadPidFile → PidAlive → [if not] proceed to Listen → WritePidFile
```

There is no advisory lock or atomic `O_CREAT|O_EXCL` on the pidfile. Two concurrent `jitenv unlock` calls (or unlock + a TUI save that triggers a respawn) can both pass `PidAlive`-returns-false, both call `SpawnDaemon`, both children try to `Listen`, and the second fails at socket-bind / pipe-create. The user sees a confusing "agent already running" or "address in use" error from one of them.

On Windows specifically, `PidAlive` uses `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)`, which returns true for any process with that PID — not just a jitenv agent. After a crashed agent leaves a stale pidfile and the PID gets reused by something else (Windows reuses PIDs aggressively), `SpawnDaemon` refuses to start with `agent already running (pid N)` even though no agent is running. Manual `rm $LOCALAPPDATA\jitenv\agent.pid` is needed to recover.

## Impact
- Spurious errors on concurrent unlock attempts.
- DoS on Windows after a crash if PID reuse hits before the user notices and removes the stale file.
- No security boundary crossed — the final state is always one running agent, no key material disclosed.

## Fix
- Replace the read-and-check pattern with an OS-level lock file:
  - Unix: `flock(LOCK_EX | LOCK_NB)` on a `.lock` file alongside the pidfile. Hold the fd open for the agent's lifetime.
  - Windows: `CreateFile` with `dwShareMode=0` and disposition `CREATE_NEW`; hold the handle for the agent's lifetime.
- On Windows, additionally verify the process image name (`QueryFullProcessImageNameW`) matches `jitenv.exe` before treating `PidAlive == true` as "agent running".

Location: `internal/agent/server.go:89-101`, `internal/agent/pidfile.go`, `daemonize_unix.go:23`, `daemonize_windows.go:44`.

## Severity
**Low** — usability defect with a small DoS angle on Windows.

Galvill:
Closes #130. See commit.

### DIFF
--- a/internal/agent/lockfile_test.go
+++ b/internal/agent/lockfile_test.go
@@ -1,0 +1,55 @@
+//go:build !windows
+
+package agent
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// TestAcquirePidLock_RejectsSecondAcquisition is the regression for
+// security #130: two concurrent unlock attempts must not both pass
+// the "already running?" check. Without an advisory lock the pidfile
+// approach has a TOCTOU window between ReadPidFile and WritePidFile.
+func TestAcquirePidLock_RejectsSecondAcquisition(t *testing.T) {
+	dir := t.TempDir()
+	path := filepath.Join(dir, "agent.pid")
+
+	first, err := acquirePidLock(path)
+	if err != nil {
+		t.Fatalf("first acquire: %v", err)
+	}
+	defer first.Close()
+
+	second, err := acquirePidLock(path)
+	if err == nil {
+		second.Close()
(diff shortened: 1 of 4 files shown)


Galvill:

PR #155, opened 2026-05-15
fix(sources/vault): cache authenticated client until lease expires (#133)

### LINKED ISSUE #133: security: Vault source re-authenticates AppRole on every Fetch — secret_id exhaustion and unnecessary network exposure
## Summary
`internal/sources/vault/vault.go:21` (package comment) and `:266` (`v.newClient(ctx)` per Fetch) — every `Fetch` call rebuilds the client and re-runs auth. For AppRole, this means the `role_id` + `secret_id` pair is sent to `/v1/auth/approle/login` on every secret fetch.

## Impact
- AppRole `secret_id` entries typically have a use-count or TTL limit. A heavy-use jitenv (many mapped commands fired per session) burns through `secret_id` lives quickly, eventually causing auth failures.
- Repeated login traffic increases the audit-log volume on Vault, making genuine anomalies harder to spot.
- The `secret_id` is transmitted on every fetch — more opportunities for `tls_skip_verify`-style misconfiguration to result in interception (see #X for the related TLS finding).
- Static-token auth has no exhaustion issue but still creates noise.

## Fix
Cache the Vault client and its token until expiry. On every fetch:
1. If `cachedToken != nil && now < expiresAt - 60s`, reuse it.
2. Otherwise authenticate, store token + `time.Add(LeaseDuration)`.

Lock around the cache for concurrent fetches. The `vaultapi` client already does token-self-renewal if asked; consider running a goroutine for that.

Location: `internal/sources/vault/vault.go:260-280` and the package-level comment.

## Severity
**Low** — operational/correctness finding with a small adjacent security improvement (less material on the wire).

Galvill:
Closes #133. See commit.

### DIFF
--- a/internal/sources/vault/vault.go
+++ b/internal/sources/vault/vault.go
@@ -26,6 +26,8 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"sync"
+	"time"
 
 	vaultapi "github.com/hashicorp/vault/api"
 	approleauth "github.com/hashicorp/vault/api/auth/approle"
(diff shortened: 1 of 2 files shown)


Galvill:

PR #156, opened 2026-05-15
fix(cli/agent): scrub inherited AWS env vars at startup (#122)

### LINKED ISSUE #122: security: AWS source inherits AWS_* env vars from unlock shell — credential chain hijack
## Summary
`internal/sources/aws/aws.go:154-155` falls through to `config.LoadDefaultConfig` (AWS SDK v2 default credential chain) when `access_key_id` is empty. The default chain honours several environment variables read from the **parent** of `jitenv unlock`:

- `AWS_SHARED_CREDENTIALS_FILE` — attacker-controlled credentials file
- `AWS_CONFIG_FILE` — attacker-controlled config (with `credential_process`, role chaining, etc.)
- `AWS_PROFILE` — selects a profile from the above
- `AWS_ROLE_ARN` + `AWS_WEB_IDENTITY_TOKEN_FILE` — IRSA / OIDC path

The agent inherits its environment from `jitenv unlock`, so anything in the unlock shell's env is in the agent's env for its entire lifetime. The legacy `profile` field was intentionally removed (commit `c06e704`) — but the default-chain path reintroduces every one of these vectors through env vars.

## Impact
An attacker who can poison the user's shell env before `jitenv unlock` runs — via a malicious `.envrc`, a hostile shell plugin, a CI runner with stale env, or a one-line append to `.bashrc` — can redirect every `aws`-source fetch to:
- An attacker-controlled `credential_process` (silent exfil of "which secrets does jitenv fetch?").
- A different AWS account entirely (substituted credentials → wrong account read → user assumes their secret is gone or different).
- A relative path that lands at `/proc/self/cwd/evil.ini` (workflows where cwd is attacker-controlled).

## Fix
- At agent startup, snapshot and **scrub** `AWS_SHARED_CREDENTIALS_FILE`, `AWS_CONFIG_FILE`, `AWS_PROFILE`, `AWS_ROLE_ARN`, `AWS_WEB_IDENTITY_TOKEN_FILE` from the agent's process env. If the user wants those honoured, surface them as explicit source-config fields (`shared_credentials_file = "..."` etc.) which go through the encrypted config and the schema.
- At minimum, log the **effective credential provider** at debug level when the default chain is used, so users can observe which path the SDK picked.
- Document this prominently in the AWS source help text.

Location: `internal/sources/aws/aws.go:140-170`, agent startup in `internal/cli/agent_internal.go`.

## Severity
**Medium-High** — depends on threat model; trivial for an attacker with any shell-startup write access.

Galvill:
Closes #122. See commit.

### DIFF
--- a/internal/cli/agent_env_scrub.go
+++ b/internal/cli/agent_env_scrub.go
@@ -1,0 +1,52 @@
+package cli
+
+import (
+	"log/slog"
+	"os"
+)
+
+// awsScrubVars lists the AWS-SDK env vars that the agent must NOT
+// inherit from the unlocking shell (security #122). Each one can
+// redirect the SDK's default credential chain at the file-path
+// level: AWS_PROFILE picks a section out of an attacker-controlled
+// credentials file; AWS_SHARED_CREDENTIALS_FILE / AWS_CONFIG_FILE
+// supply the file directly; AWS_ROLE_ARN +
+// AWS_WEB_IDENTITY_TOKEN_FILE drive IRSA / OIDC assumption against
+// an attacker-chosen token path. Users who want these honoured
+// should encode them as explicit fields under [sources.<name>.params]
+// where they round-trip through the encrypted config.
+//
+// Region-only vars (AWS_REGION / AWS_DEFAULT_REGION) are not
+// scrubbed: they don't influence which credentials are loaded.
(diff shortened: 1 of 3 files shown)


Galvill:

PR #157, opened 2026-05-15
fix(run,shim,shell): gate __JITENV_INJECTED on a per-session nonce (#120)

### LINKED ISSUE #120: security: __JITENV_INJECTED=1 pre-set in environment silently disables secret injection (no warning)
## Summary
`internal/run/run.go:52-54` and `internal/shim/shim.go:97-99` short-circuit all agent interaction when `__JITENV_INJECTED=1` is present in the env:

```go
if os.Getenv(injectedMarker) == "1" {
    return replaceProcess(abs, args, os.Environ())  // no fetch, no agent
}
```

This marker is set to `"1"` by `jitenv run` after a successful injection to prevent double-injection in execve chains (a real and legitimate concern). But the value is a constant — any process that exports `__JITENV_INJECTED=1` in the user's shell, before any mapped command runs, silently bypasses jitenv for every subsequent invocation.

## Impact
A malicious `.bashrc` append, a hostile shell plugin, a CI runner that exports stale env vars across jobs, or a package install that exports it as a "hook for tooling" can disable secret injection without any user-facing signal. The user types `kubectl get pods`, jitenv sees the marker, calls `replaceProcess` directly, and `kubectl` runs with whatever credentials are in the ambient env — often nothing, or stale ones from a previous unlock. There is no warning, no log, no exit code change.

For a tool whose purpose is "make sure the right credentials reach the right command", silent fail-open is a serious posture issue.

## Fix
Use a session-scoped nonce instead of a constant. At hook-install time (or first `jitenv run` per shell), generate a random 128-bit value and stash it in `__JITENV_SESSION` (exported by the shell hook). The injection marker becomes the same nonce; downstream `jitenv run` invocations check `__JITENV_INJECTED == __JITENV_SESSION` rather than `== "1"`. A pre-set marker with the wrong value forces a fresh injection.

Alternative: have the parent `jitenv run` `_unset_ __JITENV_INJECTED` before exec'ing the child only if it observes the marker was already set on entry **and** wasn't placed there by a recognised parent. This is hand-wavy; the nonce approach is cleaner.

Either way: when the marker is present on entry but the agent reports the path **is** mapped, log a one-line warning to stderr so the user notices.

Location: `internal/run/run.go:52-54`, `internal/shim/shim.go:97-99`.

## Severity
**Medium** — depends on the threat model; users who treat jitenv as "secrets reliably arrive" should consider this High.

Galvill:
Closes #120. See commit.

### DIFF
--- a/internal/run/run.go
+++ b/internal/run/run.go
@@ -6,7 +6,10 @@ package run
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/hex"
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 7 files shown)


Galvill:

PR #158, opened 2026-05-15
fix(agent/peer_windows): use ImpersonateNamedPipeClient (#132)

### LINKED ISSUE #132: security: Windows peer check uses GetNamedPipeClientProcessId + OpenProcess instead of ImpersonateNamedPipeClient (PID-TOCTOU)
## Summary
`internal/agent/peer_windows.go:42-54` resolves the connecting client's SID by:

1. `GetNamedPipeClientProcessId(pipe, &pid)`
2. `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, FALSE, pid)`
3. `OpenProcessToken(...)` → SID compare.

Steps 1 and 2 are not atomic against each other: if the client process exits between the two calls and Windows reuses the PID for a different process (fast on Windows under load), the SID check reads the wrong process.

The standard Windows idiom for pipe authentication is `ImpersonateNamedPipeClient(pipe)` → `OpenThreadToken(GetCurrentThread(), TOKEN_QUERY, TRUE, &tok)` → compare → `RevertToSelf()`. This binds the credential check to the pipe connection itself, not an indirectly-resolved PID.

## Impact
- The realistic failure mode is a **false rejection** (the new PID's process has a different SID → check fails → legitimate same-user client gets `unauthorized`). This is annoying but not a security vulnerability.
- A false acceptance would require the PID to be reused by a same-SID process, but the pipe's SDDL already restricts the ACL to the user's SID — any same-SID replacement would have been legitimately authorised.
- The PID approach is non-standard for pipe auth and a reviewer-bait in any future Windows-security audit.

## Fix
Replace the PID lookup with impersonation:

```go
if err := winio.ImpersonateNamedPipeClient(pipeHandle); err != nil { ... }
defer winio.RevertToSelf()
var tokenHandle windows.Token
if err := windows.OpenThreadToken(windows.CurrentThread(), windows.TOKEN_QUERY, true, &tokenHandle); err != nil { ... }
defer tokenHandle.Close()
... compare SID ...
```

`github.com/Microsoft/go-winio` exports the impersonation helpers if you don't want to call the raw Win32 API.

Location: `internal/agent/peer_windows.go:30-80`.

## Severity
**Low** — current behaviour is overly conservative (false reject) rather than permissive, but the non-standard approach is worth normalising.

Galvill:
Closes #132. See commit.

### DIFF
--- a/internal/agent/peer_windows.go
+++ b/internal/agent/peer_windows.go
@@ -5,6 +5,8 @@ package agent
 import (
 	"fmt"
 	"net"
+	"runtime"
+	"syscall"
 
 	"golang.org/x/sys/windows"
 )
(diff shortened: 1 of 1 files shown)


Galvill:

PR #159, opened 2026-05-15
fix(run,shim): drop env-string references after Windows cmd.Start (#121)

### LINKED ISSUE #121: security: Windows jitenv run keeps secrets in parent process heap for the full lifetime of the child
## Summary
Windows has no `execve`, so `internal/run/run_windows.go:78-119` and `internal/shim/shim_windows.go:84-131` spawn the child via `exec.Command`, set `cmd.Env = env` (where `env` includes fetched secret values), then `cmd.Wait()` for the child to exit.

While the child correctly receives the secrets in its Win32 environment block (and only there), the **parent** `jitenv.exe` process continues to hold the `env []string` slice — and the underlying string allocations containing secret bytes — in its Go heap for the entire lifetime of the child.

## Impact
- A long-running mapped command (`python manage.py runserver`, `terraform apply`, a dev server) keeps secrets in a second process's heap for as long as it runs.
- Any process with `PROCESS_QUERY_INFORMATION | PROCESS_VM_READ` on the parent (same user — trivially available) can `ReadProcessMemory` and recover the values.
- Windows Error Reporting / WER crash dumps, Sysinternals ProcDump, antivirus memory scans, kernel debug captures, and BSOD minidumps that hit the parent process all preserve the secrets to disk.
- On Unix, `syscall.Exec` replaces the image — there is no dangling parent — so this is a Windows-specific regression of the Unix security property.

## Fix
After `cmd.Start()` succeeds, overwrite the secret-bearing entries in `env` so the parent retains only key names:

```go
if err := cmd.Start(); err != nil { ... }
for i, kv := range env {
    if isInjected(kv) {  // entries the resolver appended
        // Replace string header so the GC reclaims the secret bytes
        env[i] = kv[:strings.IndexByte(kv, '=')+1]
    }
}
// keep cmd.Env *= env semantics: child already got its own block
```

Go strings are immutable, so the underlying byte array can't be zeroed in place — but replacing the slice element with a short string releases the original to the GC, which is the best Go allows without `unsafe`.

Better, use `windows.VirtualLock` on a private buffer for the secrets, build the child's `CreateProcessW` env block manually with `CREATE_UNICODE_ENVIRONMENT`, and never let Go strings hold the secrets.

Location: `internal/run/run_windows.go:78-119`, `internal/shim/shim_windows.go:84-131`.

## Severity
**Medium** — secrets are user-confined, but the disclosure surface is materially larger than on Unix.

Galvill:
Closes #121. See commit.

### DIFF
--- a/internal/run/run_windows.go
+++ b/internal/run/run_windows.go
@@ -15,6 +15,20 @@ import (
 	"golang.org/x/sys/windows"
 )
 
+// scrubEnvSlice replaces each entry in env with an empty string,
+// dropping the slice's reference to the original (possibly secret-
+// bearing) string. Go strings are immutable; this can't overwrite
+// the heap bytes directly, but it removes the slice from the GC
+// reachability graph for those strings so a future collection can
+// reclaim them. Used on the Windows replaceProcess path after
+// cmd.Start has copied the env into the child's Win32 block
+// (security #121).
+func scrubEnvSlice(env []string) {
+	for i := range env {
+		env[i] = ""
+	}
+}
+
 // scriptInterpreter returns the executable + leading args needed to run
 // realPath when it isn't a directly-launchable Win32 binary. CreateProcess
 // only accepts PE files (.exe, .com) and the legacy DOS-style .bat/.cmd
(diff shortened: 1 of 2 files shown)


Galvill:

PR #160, opened 2026-05-15
fix(agent,cli): hand master key over stdin on Windows, not via cmdline handle (#128)

### LINKED ISSUE #128: security: Windows --key-handle=<hex> exposes inheritable handle value on child command line
## Summary
`internal/agent/daemonize_windows.go:85-91` passes the inherited pipe handle to the agent child as `--key-handle=<hex>` on the command line. The hex value is the kernel handle ID of the anonymous-pipe read end in the child's handle table. On Windows, command lines are readable by any process running as the same user via WMI / `NtQueryInformationProcess` / Process Explorer.

The master-key bytes themselves are not on the command line — only the handle is — but a same-user attacker can:
1. Enumerate the child's command line (no special privilege needed, same user).
2. `OpenProcess(PROCESS_DUP_HANDLE, FALSE, childPid)` to get a handle to the child.
3. `DuplicateHandle(hChild, <hex>, hSelf, ...)` to copy the read end of the pipe into their own process.
4. `ReadFile` on the duplicated handle — reads whatever key bytes are still in the pipe buffer.

The race window is small: the parent writes the key immediately after `cmd.Start()` and the child closes the handle after reading. Both happen in tens of milliseconds. Same-user privilege is already required, but `DuplicateHandle` is fast.

## Impact
Local same-user privilege escalation from "process running as user" to "agent's master key", winning a small race. Not "any process can read the key" — but tighter than the Unix `fd=3` analogue, which never appears in any process-visible string.

## Fix
- Instead of passing the handle value, always inherit the read end at a **fixed, well-known** handle slot. `STARTUPINFOEX` with `PROC_THREAD_ATTRIBUTE_HANDLE_LIST` already restricts inheritance — pair that with `bInheritHandles=TRUE` on a specific extra handle that the child looks up by index/name rather than by hex value.
- Or pass the key over an environment variable set only in `cmd.Env` (and not the parent's env), clearing it from the parent immediately after `CreateProcess`. Env-on-cmdline is still a Windows footgun, so prefer the fixed-slot approach.

Document the trade-off in `internal/agent/daemonize_key_windows.go` if the current design is intentional.

Location: `internal/agent/daemonize_windows.go:85-91`, `internal/cli/agent_key_windows.go`.

## Severity
**Medium** — narrow race, same-user privilege required, but the handle-value disclosure is unique to the Windows path and worth eliminating.

Galvill:
Closes #128. See commit.

### DIFF
--- a/internal/agent/daemonize_key_windows.go
+++ b/internal/agent/daemonize_key_windows.go
@@ -11,13 +11,9 @@ import (
 )
 
 // ReadKeyFromHandle reads exactly KeyLen bytes from the given Windows
-// handle. The handle is expected to be the read end of an anonymous pipe
-// that the parent SpawnDaemon inherited into this process via
-// SysProcAttr.AdditionalInheritedHandles. The caller is responsible for
-// zeroing the returned slice.
-//
-// Mirrors ReadKeyFromFd on Unix; lives in agent because the daemon side
-// is what cracks open the inherited handle.
+// handle. Retained for the legacy --key-handle=<hex> path (test
+// fixtures); production SpawnDaemon now hands the key over stdin
+// (security #128) — see ReadKeyFromReader.
 func ReadKeyFromHandle(h syscall.Handle) ([]byte, error) {
 	f := os.NewFile(uintptr(h), "key-pipe")
 	if f == nil {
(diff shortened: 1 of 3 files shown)


Galvill:

PR #161, opened 2026-05-15
fix(crypto,cli): mlock/VirtualLock the master key buffer (#127)

### LINKED ISSUE #127: security: no mlock / VirtualLock on master key or decrypted secrets — pages may be swapped to disk
## Summary
The master KEK (`key []byte`, lifetime: entire agent process) and every decrypted source-param / bag-value live in normal Go heap memory. Neither `unix.Mlock` (Linux/macOS) nor `windows.VirtualLock` is invoked.

On systems with swap enabled (the default on Linux desktops, all of Windows), under memory pressure the OS can page these allocations to disk. The 64 MiB Argon2id working buffer (`internal/crypto/kdf.go:43`) is even larger and more likely to be selected for eviction.

## Impact
- Cold-boot / disk-forensics recovery of the swap partition or pagefile reveals key bytes and decrypted secrets.
- On cloud VMs, the hypervisor controls disk image storage; key material in the guest's swap is exposed to the cloud operator.
- Core dumps inherit the same risk: by default Linux includes anonymous memory in cores.

## Fix
1. After `ReadKeyFromFd` returns the key bytes, call `unix.Mlock(key)` / `windows.VirtualLock(...)`. Bail loudly if the call fails (mlock has an RLIMIT_MEMLOCK ceiling — fall back to a printed warning and proceed).
2. Call `unix.Madvise(MADV_DONTDUMP, key)` on Linux so cores exclude the page.
3. Apply the same lock to decrypted secret buffers — easier once the zero-on-lock refactor lands.
4. For the Argon2id 64-MiB working buffer, the `argon2` package allocates a fresh buffer per call — locking that is harder; consider replacing with a wrapper that pre-allocates an mlocked region.

Location: `internal/cli/agent_internal.go:35-75`, `internal/agent/daemonize_key_windows.go`, `internal/crypto/kdf.go`.

## Severity
**Medium** — defence against post-mortem forensics; not exploitable while the agent is running.

Galvill:
Closes #127. See commit.

### DIFF
--- a/internal/cli/agent_internal.go
+++ b/internal/cli/agent_internal.go
@@ -37,6 +37,7 @@ func newAgentInternalCmd() *cobra.Command {
 				return fmt.Errorf("read key: %w", err)
 			}
 			defer zeroBytes(key)
+			defer lockKey(key)()
 			slog.Info("agent starting", "config", cfgArg, "idle", idle.String())
 
 			loadAndBuild := func() (agent.Resolver, error) {
--- a/internal/cli/config.go
+++ b/internal/cli/config.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"fmt"
+	"log/slog"
 
 	"github.com/BurntSushi/toml"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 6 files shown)


Galvill:

PR #162, opened 2026-05-15
fix(agent,sources/local): wipe cached plaintext secrets on Shutdown (#125)

### LINKED ISSUE #125: security: decrypted secrets and master key are not zeroed on OpLock / agent shutdown
## Summary
`internal/cli/agent_internal.go:39` correctly does `defer zeroBytes(key)` for the raw KDF output — but the **decrypted** secrets reachable through the resolver are not zeroed when the agent receives `OpLock` or shuts down.

`internal/agent/resolver.go` holds `cfg.Secrets` (a `map[string]map[string]string` of plaintext bag values) via `SetBags`. Source params decrypted by `DecryptStringsInPlace` (`internal/config/decrypt.go`) live in `cfg.Sources[*].Params` as plaintext `string` values. None of these are explicitly cleared on lock; they wait for GC.

Compounding this, `internal/crypto/envelope.go:38` does `return string(pt), nil` — once a secret is a Go string, the underlying bytes are immutable and unzeroable from Go.

## Impact
- A core dump of the agent immediately after `jitenv lock` still contains every plaintext secret until GC overwrites the heap (which may never happen for a process that exits soon after lock).
- `/proc/<pid>/mem` read by root, swap pages on disk, hypervisor memory snapshots, and Windows crash dumps all preserve the cleartext for hours to forever.
- The threat model already includes "memory dump of agent → leak" — but `defer zeroBytes(key)` shows the project's stated intent to minimise that window. The secrets themselves are the more valuable target than the KEK, and they're left in the clear.

## Fix
- For the `local` source (`internal/sources/local/local.go`), store bags as `map[string][]byte` rather than `map[string]string` and overwrite the byte slices on `Close`/lock.
- Threaded through: `DecryptStringsInPlace` should return `[]byte` for sensitive fields (or expose a separate `DecryptBytes`-into-`*[]byte` API) so the agent can avoid creating immutable string copies in the first place.
- Add a `(*Agent).Shutdown` step that calls a `Zero()` method on the resolver, which propagates to each source plugin to wipe whatever it cached.
- For the master key on Windows: hold it in `windows.VirtualLock`'d memory and overwrite before unlock.

This is a large refactor; the minimum acceptable interim fix is to (1) overwrite `cfg.Secrets` values to empty strings on `Shutdown` and (2) clear `cfg.Sources[*].Params` string-typed values the same way. These overwrites only release the original heap allocations to GC — they don't zero them — but reduce the residency window.

Location: `internal/agent/server.go:135-144`, `internal/agent/resolver.go`, `internal/crypto/envelope.go:38`.

## Severity
**Medium** — known industry-wide limitation of Go-based secret tools, but worth tightening given jitenv's stated key-hygiene discipline.

Galvill:
Closes #125. See commit.

### DIFF
--- a/internal/agent/resolver.go
+++ b/internal/agent/resolver.go
@@ -25,6 +25,36 @@ type bagSink interface {
 	SetBags(map[string]map[string]string)
 }
 
+// wiper is implemented by sources that cache decrypted secret
+// material (the local source's bag map, future caches for remote
+// sources) and need a hook to drop those references on shutdown
+// (security #125). Optional — sources that hold no plaintext
+// secrets don't implement it.
+type wiper interface {
+	Wipe()
+}
+
+// Wipe walks every source the resolver owns and asks it to release
+// any cached plaintext. Then it drops the resolver's own reference
+// to the per-source config map (which still holds the decrypted
+// param strings post-DecryptInPlace). Best-effort: Go strings are
+// immutable, so this can't overwrite the heap bytes — it removes
+// the live reference chain so a future GC can reclaim the memory.
+func (r *resolver) Wipe() {
+	for _, s := range r.sources {
+		if w, ok := s.(wiper); ok {
+			w.Wipe()
+		}
+	}
+	for name, sc := range r.cfg {
+		for k := range sc.Params {
+			sc.Params[k] = ""
+		}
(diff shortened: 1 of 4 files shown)


Galvill:

PR #167, opened 2026-05-19
fix(shell,cli): warn on unsupported shells at unlock (#164)

### LINKED ISSUE #164: jitenv unlock: warn when the current shell isn't supported (fish, tcsh, ksh, dash, cmd.exe, …)
## Severity
medium (silent failure for any user not on bash/zsh/pwsh — they unlock, see \"agent started\", and never realise the hook will never load in their shell)

## Where
- `internal/shell/install.go:17-32` — `DetectShell()` collapses \"unsupported shell\" and \"\$SHELL unset\" into the same empty-string return
- `internal/cli/unlock.go:88` calls `warnIfHookMissing`, which at `unlock.go:107-109` early-returns when `st.Shell == \"\"` — so unsupported shells get no warning at all
- `internal/shell/install.go:129-150` — `CurrentStatus()` propagates the same shape; nothing downstream can tell \"unknown\" from \"unsupported\"

## Problem

`jitenv unlock` from a supported shell prints any relevant hook-status notice (e.g. `note: shell hook not installed in ~/.zshrc — run jitenv hook install …`). From an *unsupported* shell — fish, tcsh, ksh, dash, busybox sh, nu, xonsh, sh-only POSIX, cmd.exe — `jitenv unlock` silently:

```
$ jitenv unlock
unlock passphrase:
agent started (socket: /run/user/1000/jitenv/agent.sock)
```

No hint that the user just unlocked the agent into a session where the shell hook will never run. Path-mapped commands won't auto-inject, `cwd_glob` wrapper symlinks won't activate, the agent-down countdown UX in `internal/agentwarn/agentwarn.go` never fires because nothing routes through `jitenv run`. The agent is up but functionally invisible.

The cause is the way `DetectShell()` reports failure:

```go
// internal/shell/install.go:17-32
func DetectShell() string {
    if runtime.GOOS == \"windows\" { return \"powershell\" }
    sh := os.Getenv(\"SHELL\")
    if sh == \"\" { return \"\" }
    switch filepath.Base(sh) {
    case \"bash\": return \"bash\"
    case \"zsh\":  return \"zsh\"
    }
    return \"\"
}
```

Both \"shell envvar absent\" and \"shell envvar names something unsupported\" return `\"\"`. The consumer at `unlock.go:107-109` then can't differentiate:

```go
st, err := shell.CurrentStatus()
if err != nil || st.Shell == \"\" {
    return
}
```

## Repro

```sh
# In any non-bash/zsh shell on Linux/macOS.
fish -c 'jitenv unlock'        # or tcsh, ksh, dash, …
# → \"agent started …\" with no hook-status warning.
```

## Desired

When `jitenv unlock` detects an unsupported shell, surface a yellow notice in the same family as the existing hook-missing notices:

```
agent started (socket: /run/user/1000/jitenv/agent.sock)
note: detected shell \"fish\" (from $SHELL=/usr/bin/fish) is not supported by jitenv.
      Only bash, zsh, and PowerShell 7+ have shell hooks. The agent will run, but
      path-mapped commands won't auto-inject env vars in this shell. See
      https://github.com/Galvill/jitenv#supported-shells.
```

For the \"$SHELL is unset / unreadable\" case (no useful name to print), keep the current silent behaviour — there's nothing actionable to say.

On Windows, today `DetectShell()` unconditionally returns `\"powershell\"`. That's an existing optimism — `cmd.exe` and Windows PowerShell 5.x users get no warning. Worth covering as part of this fix: inspect the parent process (`syscalls.GetParent` / `gopsutil` style) to see whether it's actually `pwsh.exe`, and warn when it isn't. Out-of-scope-but-related is whether `$SHELL` even being the right input on Unix — it names the *login* shell, not necessarily the current interactive one. Parent-process inspection is more accurate everywhere; doing it just on Windows is the cheap first move because it's the only place an obviously-wrong default exists today.

### 1. Make `DetectShell` distinguish unknown from unsupported

```go
// New signature: returns the canonical name when supported, the raw
// detected basename when unsupported, and empty when nothing was
// detectable. Callers that just want supported-only can keep using
// DetectShell() unchanged (returns \"\" for both unknown and unsupported).
func DetectShellDetailed() (canonical, raw, source string) {
    if runtime.GOOS == \"windows\" {
        return \"powershell\", \"\", \"runtime.GOOS==windows\"
    }
    sh := os.Getenv(\"SHELL\")
    if sh == \"\" {
        return \"\", \"\", \"\"
    }
    base := filepath.Base(sh)
    switch base {
    case \"bash\":
        return \"bash\", base, sh
    case \"zsh\":
        return \"zsh\", base, sh
    }
    return \"\", base, sh
}

func DetectShell() string {
    canon, _, _ := DetectShellDetailed()
    return canon
}
```

`canonical == \"\" && raw != \"\"` → unsupported, name to print is `raw`, originating value is `source`.
`canonical == \"\" && raw == \"\"` → truly unknown, stay silent.

### 2. Extend `Status` and `CurrentStatus`

```go
type Status struct {
    Shell        string // supported canonical name; empty when unsupported/unknown
    Unsupported  string // populated only when an unsupported shell was detected
    Source       string // $SHELL value (or equivalent) for the warning body
    // …existing fields…
}

func CurrentStatus() (Status, error) {
    canon, raw, source := DetectShellDetailed()
    if canon == \"\" {
        return Status{Unsupported: raw, Source: source}, nil
    }
    // …existing path…
}
```

### 3. Branch the unlock notice

```go
// internal/cli/unlock.go warnIfHookMissing
st, err := shell.CurrentStatus()
if err != nil { return }

if st.Shell == \"\" {
    if st.Unsupported != \"\" {
        fmt.Fprintf(w,
            \"%snote:%s detected shell %q (from $SHELL=%q) is not supported by jitenv. \"+
            \"Only bash, zsh, and PowerShell 7+ have shell hooks. The agent is running, \"+
            \"but path-mapped commands won't auto-inject env vars in this shell.\\n\",
            yellow, reset, st.Unsupported, st.Source)
    }
    return // silent on truly-unknown
}
// …existing Installed/LoginSources checks…
```

### 4. (Bonus, follow-up scope) Parent-process check on Windows

Detect whether the calling process's parent is `pwsh.exe` vs `powershell.exe` (5.x) vs `cmd.exe`. Print the same unsupported-shell warning when it's not pwsh 7+. Can use the standard-library `os.Getppid` plus a Windows-only PID→name lookup via `golang.org/x/sys/windows`; or rely on the `PSModulePath` heuristic (pwsh 7+ adds a distinct entry).

This part is genuinely platform-y enough to file as its own issue if scope is tight — \"warn on cmd.exe / Windows PowerShell 5.x\" with this issue's plumbing already in place. I'd flag it here so the `Status.Unsupported` field gets designed with that future caller in mind, but not block the Linux/macOS fix on it.

## Notes

- The TUI also calls `shell.CurrentStatus()` (see `internal/tui/settings_screen.go`) to render the shell-hook section. Once `Status.Unsupported` exists, the settings screen should render a parallel notice — \"Hook not available for this shell\" instead of empty/missing state — but that's a follow-up; the unlock path is where users *first* encounter the silent-failure mode.
- `internal/cli/hook.go` (the `jitenv hook install` / `jitenv hook <shell>` subcommands) is the other place that should grow an explicit \"unsupported shell\" error message. Today `jitenv hook install` with no `--shell` flag in fish would call `DetectShell()`, get `\"\"`, and presumably fall through to a generic \"can't determine shell\" error. Worth a once-over for consistent wording.
- The supported-shells list is currently maintained implicitly across `DetectShell`, `RcPath`, `HookLine`, and `installPathsFor`. The fix is a good moment to add a single `SupportedShells() []string` helper that returns `[\"bash\", \"zsh\", \"powershell\"]` so the warning body can include it instead of being hardcoded.

## Test plan

- `SHELL=/usr/bin/fish jitenv unlock` → unlock succeeds, unsupported-shell warning printed.
- `SHELL=/bin/sh jitenv unlock` → same.
- `unset SHELL; jitenv unlock` → no warning (truly unknown).
- `SHELL=/bin/bash jitenv unlock` → existing hook-missing / login-chain warnings unchanged.
- A unit test on `DetectShellDetailed` covering: bash, zsh, fish, sh, empty, Windows.

## Related

- `agentwarn.WarnAndWait` (#64) only fires once a command routes through `jitenv run`. In unsupported shells nothing ever routes through `jitenv run`, so #64's whole UX is bypassed. The unlock warning is the only signal a non-bash/zsh/pwsh user gets that their setup is incomplete.
- #107 (PowerShell hook debug visibility) — adjacent area; if the parent-process check from the bonus section above lands, those users get a clearer error than \"why is `JITENV_HOOK_DEBUG` doing nothing\".

Galvill:
Closes #164. See commit.

### DIFF
--- a/internal/cli/unlock.go
+++ b/internal/cli/unlock.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -94,21 +95,38 @@ func newUnlockCmd() *cobra.Command {
 }
 
 // warnIfHookMissing prints a yellow notice to w when the shell hook
-// isn't fully wired up. Two failure modes are flagged:
+// isn't fully wired up. Three failure modes are flagged:
+//   - the user is in an unsupported shell (fish / dash / ksh / …) so
+//     the hook will never load — issue #164;
 //   - the eval line is not in the user's interactive rc file at all;
-//   - the eval line is in ~/.bashrc but bash login shells don't end up
-//     sourcing ~/.bashrc, so the hook only fires in interactive shells
-//     (this is the situation that hides the agent-down warning when
-//     the user opens a new terminal as a login shell).
+//   - the eval line is in ~/.bashrc but bash login shells don't end
+//     up sourcing ~/.bashrc, so the hook only fires in interactive
+//     shells.
 func warnIfHookMissing(w interface {
 	Write(p []byte) (n int, err error)
 }) {
 	st, err := shell.CurrentStatus()
-	if err != nil || st.Shell == "" {
+	if err != nil {
 		return
 	}
 	const yellow = "\033[33m"
 	const reset = "\033[0m"
+
+	if st.Shell == "" {
+		// Truly-unknown shell ($SHELL unset) stays silent — nothing
+		// actionable to say. Unsupported shell with a name gets a
+		// dedicated warning so the user knows the agent is running
+		// but the hook will never auto-inject (#164).
+		if st.Unsupported != "" {
+			fmt.Fprintf(w,
+				"%snote:%s detected shell %q (from $SHELL=%q) is not supported by jitenv. "+
+					"Only %s have hooks; the agent is running, but path-mapped commands "+
+					"won't auto-inject env vars in this shell.\n",
+				yellow, reset, st.Unsupported, st.Source,
+				strings.Join(shell.SupportedShells(), ", "))
+		}
+		return
+	}
 	if !st.Installed {
 		fmt.Fprintf(w,
 			"%snote:%s shell hook not installed in %s — run `jitenv hook install` "+
--- a/internal/shell/install.go
+++ b/internal/shell/install.go
@@ -10,25 +10,59 @@ import (
 	"strings"
 )
 
-// DetectShell returns "bash" / "zsh" / "powershell", or "" if the
-// current shell isn't supported. On Windows $SHELL is normally unset,
-// so we assume PowerShell 7+ (the only Windows shell jitenv supports —
-// see issue #39 for the cmd.exe / PowerShell 5.x decision).
+// SupportedShells lists the canonical names jitenv has hook snippets
+// for. Used by user-facing warnings so the wording stays in sync with
+// the actual matrix.
+func SupportedShells() []string {
+	return []string{"bash", "zsh", "powershell"}
+}
+
+// DetectShell returns "bash" / "zsh" / "powershell" when the current
+// shell is one jitenv supports, or "" otherwise. "" collapses both
+// "$SHELL unset / unreadable" and "named-but-unsupported" cases; call
+// DetectShellDetailed when those need to be distinguished (e.g. to
+// warn an unsupported-shell user that their hook will never load).
 func DetectShell() string {
+	canonical, _, _ := DetectShellDetailed()
+	return canonical
+}
+
+// DetectShellDetailed reports what jitenv knows about the calling
+// shell (#164):
+//
+//   - canonical: one of bash/zsh/powershell when supported, "" otherwise.
+//   - raw: the basename of $SHELL when present, "" when the env var
+//     is unset (or, on Windows, "" because no detection happened).
+//   - source: the full $SHELL value (or other origin token) used to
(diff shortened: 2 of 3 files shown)


Galvill:

PR #168, opened 2026-05-19
fix(tui): pass left/right arrows to focused text input (#106)

### LINKED ISSUE #106: TUI: left/right arrow keys swallowed by inputScreen and key/value editor — can't move cursor inside the field
## Severity
medium (UX papercut — affects every single-input prompt in the TUI: bag/key creation and rename, mapping edit, ARN entry, command list entries, settings input, etc. To fix a typo mid-field you currently have to backspace everything to the typo and retype the tail.)

## Where
- `internal/tui/input_screen.go:112-121` — `case \"left\":` / `case \"right\":` unconditionally `return` regardless of focus
- `internal/tui/secrets_screen.go:492-501` — same shape on the key/value editor

For reference, `internal/tui/sources_screen.go:463-472` already gets this right and is a good template (see below).

## Problem

In any inputScreen, pressing ← or → while the text field is focused does nothing. The bubbletea `textinput.Model` would normally move the cursor on left/right, but the wrapping screen's `Update` intercepts both keys, treats them as button-row navigation, and returns before delegating to the textinput:

```go
// internal/tui/input_screen.go:112-121
case \"left\":
    if s.btnFocus > 0 {
        s.btnFocus--
    }
    return s, nil
case \"right\":
    if s.btnFocus >= 0 && s.btnFocus < len(s.buttons)-1 {
        s.btnFocus++
    }
    return s, nil
```

When `btnFocus < 0` (i.e. the input is focused), both branches still match and `return s, nil` runs. The end-of-method delegator that *would* forward arbitrary keys to the input:

```go
// internal/tui/input_screen.go:160-164
if s.btnFocus < 0 {
    var cmd tea.Cmd
    s.input, cmd = s.input.Update(msg)
    return s, cmd
}
```

…never gets a chance because the switch already returned.

`internal/tui/secrets_screen.go:492-501` has the exact same shape and is broken in the same way for the key/value text fields.

### Repro

1. `jitenv tui` → Local Sources → `< Create New Bag >`.
2. Type `myapp-prdo` (intentional typo).
3. Press ← four times. → cursor doesn't move.
4. Press Backspace four times. → end-of-string deletes; only way to fix the typo.

Same in: rename bag, add/rename key, add/edit mapping path, add ARN, etc. — anything that pushes an `inputScreen`.

## Why `sources_screen.go` is fine

The form-bearing source-edit screen does it correctly: it only returns when there's something to do for the button row, and otherwise falls through to the form's `Update` (which forwards to the focused textinput):

```go
// internal/tui/sources_screen.go:463-472 — pattern to copy
case \"left\":
    if s.btnFocus > 0 {
        s.btnFocus--
        return s, nil
    }
case \"right\":
    if s.btnFocus >= 0 && s.btnFocus < len(s.buttons)-1 {
        s.btnFocus++
        return s, nil
    }
}
// …then below the switch:
if s.btnFocus < 0 && s.form != nil {
    cmd := s.form.Update(msg)
    return s, cmd
}
```

Note the conditional `return` *inside* the `if` — the case only short-circuits when the button row actually wants the key. Otherwise control falls through to the textinput delegate.

The multi-field `form` widget (`internal/tui/form.go`) is also fine: it never intercepts left/right, so left/right reach the textinput inside `form.Update`.

## Suggested fix

Mirror the `sources_screen.go` shape in the two broken places:

### `internal/tui/input_screen.go`

```go
case \"left\":
    if s.btnFocus > 0 {
        s.btnFocus--
        return s, nil
    }
    // input focused or already on first button — let the input handle ←
case \"right\":
    if s.btnFocus >= 0 && s.btnFocus < len(s.buttons)-1 {
        s.btnFocus++
        return s, nil
    }
    // input focused or already on last button — let the input handle →
```

Combined with the existing end-of-method delegator (`if s.btnFocus < 0 { s.input.Update(msg) … }`), left/right now reach the textinput when input is focused.

### `internal/tui/secrets_screen.go`

Same change at lines 492-501. The end-of-method already delegates to `s.keyIn` / `s.valIn` based on `s.field`, so the textinput will get left/right once the case stops short-circuiting.

### Edge case worth a thought

When the cursor is at the very start of the input and the user presses ← again, the textinput will absorb the keystroke as a no-op. If we want ← at column 0 to jump focus *out* of the input to a button row (some TUIs do this), that would be a follow-up. The minimum viable fix is to stop swallowing the key.

## Test

Add a small Update-driven test that:

1. Constructs an `inputScreen` with initial value `\"abc\"`.
2. Sends `tea.KeyMsg{Type: tea.KeyLeft}` once.
3. Sends a rune `'X'`.
4. Asserts `s.input.Value() == \"abXc\"`.

Same shape for the secrets editor.

Galvill:
Closes #106. See commit.

### DIFF
--- a/internal/tui/input_screen.go
+++ b/internal/tui/input_screen.go
@@ -110,15 +110,19 @@ func (s *inputScreen) Update(msg tea.Msg) (screen, tea.Cmd) {
 			}
 			return s, nil
 		case "left":
+			// Only consume when actually moving button focus. When
+			// the text field is focused (btnFocus < 0) or already at
+			// the leftmost button, fall through to the textinput so
+			// the user can move the cursor inside the field (#106).
 			if s.btnFocus > 0 {
 				s.btnFocus--
+				return s, nil
 			}
-			return s, nil
 		case "right":
 			if s.btnFocus >= 0 && s.btnFocus < len(s.buttons)-1 {
 				s.btnFocus++
+				return s, nil
 			}
-			return s, nil
 		case "down":
 			if s.btnFocus < 0 {
 				s.btnFocus = 0
--- a/internal/tui/secrets_screen.go
+++ b/internal/tui/secrets_screen.go
@@ -490,15 +490,19 @@ func (s *kvEditScreen) Update(msg tea.Msg) (screen, tea.Cmd) {
 			}
 			return s, nil
 		case "left":
+			// Only consume when actually moving button focus. When
+			// the key/value text inputs are focused or we're already
+			// at the leftmost button, fall through so the user can
+			// move the cursor inside the field (#106).
 			if s.btnFocus > 0 {
 				s.btnFocus--
+				return s, nil
 			}
-			return s, nil
 		case "right":
 			if s.btnFocus >= 0 && s.btnFocus < len(s.buttons)-1 {
 				s.btnFocus++
+				return s, nil
 			}
-			return s, nil
 		case "enter":
 			if s.btnFocus < 0 {
 				return s, s.commit()


Galvill:

PR #169, opened 2026-05-19
fix(tui/mappings): prompt Resume/Discard on Esc for partial mappings (#163)

### LINKED ISSUE #163: TUI: Esc on a partially-filled new mapping silently saves the half-mapping — prompt Resume / Discard instead
## Severity
medium (data-shape correctness — users can accidentally save a mapping that has a target but no variables, or a cwd_glob with no commands; the saved mapping is inert and accumulates as cruft)

## Where
- `internal/tui/mappings_screen.go:283-290` — the Esc handler on `mappingFormScreen`
- `internal/tui/mappings_screen.go:256-259` — `isEmpty()` only catches *fully* blank mappings; nothing checks the partial-but-not-complete case
- `internal/tui/confirm.go:20` — `newConfirmScreen` is already the right primitive for the proposed prompt

## Problem

The mapping form has a three-step (or four-step for `cwd`) flow: pick kind → set target → pick variables (and for `cwd`, set commands). The current Esc behaviour at `mappings_screen.go:283-290`:

```go
case \"esc\":
    if s.creating && s.isEmpty() {
        // …drop the stub mapping…
        return s, tea.Sequence(emit(popMsg{}), emit(mappingChangedMsg{}))
    }
    return s, emit(popMsg{})
```

`isEmpty()` is defined as *every* slot blank:

```go
func (s *mappingFormScreen) isEmpty() bool {
    mp := s.mp()
    return mp != nil && mp.Path == \"\" && mp.Glob == \"\" && mp.CwdGlob == \"\" && len(mp.Vars) == 0
}
```

So the user gets one of two outcomes on Esc from `< Create New Mapping >`:

- **Touched nothing** → stub silently removed. Good.
- **Touched something but didn't finish** → the half-configured mapping is kept and appears in the mappings list as `(empty)`, `/some/path  (0 vars)`, `cwd:~/work/acme` (no commands), etc. It will be persisted on the next save and never injects anything when matched.

### Concrete failure shapes I can produce in five seconds each

1. Pick kind `path` → type a target → hit Esc. Mapping is saved with no variables. Inert.
2. Pick kind `cwd` → type a cwd_glob → tick a couple of variables → hit Esc. No `commands` set, so the mapping wraps zero binaries. Inert.
3. Pick kind `glob` and hit Esc immediately. Mapping survives with an empty glob string — `isEmpty()` returns true (no fields filled) so it's dropped. Fine *today*, but if the user then changes `pendingKind` semantics to count as \"touched\", the same shape leaks too.

For (1) and (2) the user has to notice the inert row in the mappings list, open the menu, and Delete it manually. New users probably won't — they'll just wonder why their env vars don't appear.

## Desired

On Esc from a new mapping that is **partially configured** (something filled, but not enough to be a useful mapping), push a small confirm modal:

```
                ┌─────────────────────────────────────────┐
                │ Mapping is incomplete:                  │
                │   target set, no variables added        │
                │                                         │
                │   [ Resume configuring ]  [ Discard ]   │
                └─────────────────────────────────────────┘
```

- **Resume** → close the modal, stay on the form, focus the first unfilled row.
- **Discard** → drop the stub mapping like the current empty-on-Esc path does, return to the mappings list.
- Esc inside the modal = Resume (least destructive).

For *complete* mappings (saveable as-is) Esc continues to pop normally with no prompt — finished configurations don't need the prompt. The current fully-empty path also stays silent.

Same prompt fires when editing an existing mapping if the user *deleted* fields and the mapping is now partial — though this is a smaller win since the user explicitly removed things.

### 1. Replace `isEmpty()` with a tri-state predicate

```go
type mappingFillState int

const (
    mappingFillEmpty    mappingFillState = iota // every field blank
    mappingFillPartial                          // some but not all required fields set
    mappingFillComplete                         // saveable
)

func (s *mappingFormScreen) fillState() (mappingFillState, string) {
    mp := s.mp()
    if mp == nil { return mappingFillEmpty, \"\" }

    targetSet := mp.Path != \"\" || mp.Glob != \"\" || mp.CwdGlob != \"\"
    varsSet   := len(mp.Vars) > 0
    cmdsSet   := len(mp.Commands) > 0

    switch {
    case !targetSet && !varsSet:
        return mappingFillEmpty, \"\"
    case mp.CwdGlob != \"\" && (!varsSet || !cmdsSet):
        switch {
        case !targetSet:               return mappingFillPartial, \"no target set\"
        case !cmdsSet && !varsSet:     return mappingFillPartial, \"no commands or variables added\"
        case !cmdsSet:                 return mappingFillPartial, \"no commands added (cwd mappings need at least one)\"
        case !varsSet:                 return mappingFillPartial, \"no variables added\"
        }
    case !targetSet:
        return mappingFillPartial, \"variables set, no target\"
    case !varsSet:
        return mappingFillPartial, \"target set, no variables added\"
    }
    return mappingFillComplete, \"\"
}
```

The second return is the human-readable reason for the prompt body.

### 2. Wire the Esc handler

```go
case \"esc\":
    state, reason := s.fillState()
    switch state {
    case mappingFillEmpty:
        if s.creating { /* existing silent-drop branch */ }
        return s, emit(popMsg{})

    case mappingFillPartial:
        if !s.creating {
            // Editing an existing mapping that's been partially cleared — same prompt,
            // different Discard semantics (revert to last-saved state instead of remove).
            return s, emit(pushMsg{s: newConfirmScreen(s.root,
                \"Mapping is now incomplete (\" + reason + \"). Discard changes?\",
                s.confirmPartialExit, \"Resume\", \"Discard\")})
        }
        return s, emit(pushMsg{s: newConfirmScreen(s.root,
            \"Mapping is incomplete (\" + reason + \").\",
            s.confirmPartialExit, \"Resume configuring\", \"Discard\")})

    case mappingFillComplete:
        return s, emit(popMsg{})
    }
```

`confirmPartialExit` is a closure that pops the modal on Resume and runs the existing empty-drop logic on Discard.

### 3. Bonus: focus the first unfilled row on Resume

`s.cursor = firstUnfilledRow()` before returning. Cheap polish that makes the prompt actively useful instead of just a speed bump.

## Notes / edge cases

- `Save` (`Ctrl-S` from the form's parent menu) already saves whatever's in `cfg.Mappings`, including partial mappings. The save path doesn't validate. A complementary follow-up issue could `Config.Validate()` partial mappings out of the saved file too, but it's not required for this fix — the Esc prompt prevents 95% of accidental partials, and `Save` is at least an explicit user action.
- Saving from a different screen *while the form is open* (Ctrl-S elsewhere, multi-screen stack) is fine because the form mutates `cfg.Mappings` in place on every popup commit; the half-mapping is already persisted by then. The new prompt only catches Esc, so a user who Ctrl-S's their way out won't be guarded — that's the same as today.
- Don't show the prompt when the form was opened in `creating=false` mode and the user only navigated through rows without changing anything. The Esc handler already returns `popMsg{}` straight through for `fillState() == mappingFillComplete` in that path.

## Test plan

- Open `< Create New Mapping >`, pick kind `path`, type target, Esc → prompt appears with \"target set, no variables added\". Pick Resume → focus is on the variables row. Pick Discard → return to list with no new mapping.
- Same flow but with `cwd` kind plus target plus variables ticked, no commands → prompt says \"no commands added (cwd mappings need at least one)\".
- Open `< Create New Mapping >`, Esc immediately → no prompt; stub is silently dropped.
- Open an existing mapping, navigate without editing, Esc → no prompt.
- Open an existing mapping, clear the variables, Esc → prompt with discard-changes semantics.

## Related

- #108 (mapping-kind picker has no description) — adjacent UX polish in the same form.
- Anywhere else the TUI lets a half-built object survive Esc is worth auditing on the back of this. Bag with no keys, source with no `region` or no ARNs, etc. — out of scope here but the same confirm-on-incomplete-Esc pattern is reusable.

Galvill:
Closes #163. See commit.

### DIFF
--- a/internal/tui/mappings_screen.go
+++ b/internal/tui/mappings_screen.go
@@ -258,6 +258,43 @@ func (s *mappingFormScreen) isEmpty() bool {
 	return mp != nil && mp.Path == "" && mp.Glob == "" && mp.CwdGlob == "" && len(mp.Vars) == 0
 }
 
+// isPartial reports whether the in-progress mapping has been touched
+// but isn't yet complete enough to be saved. A complete mapping needs
+// (a) a target (Path / Glob / CwdGlob set), (b) at least one VarRef,
+// and (c) for cwd mappings, at least one entry in Commands. Anything
+// in between — target without vars, cwd_glob without commands, vars
+// without target — is a partial mapping and should prompt the user
+// before Esc silently saves it as cruft (#163).
+func (s *mappingFormScreen) isPartial() bool {
+	mp := s.mp()
+	if mp == nil {
+		return false
+	}
+	if mp.Path == "" && mp.Glob == "" && mp.CwdGlob == "" && len(mp.Vars) == 0 {
+		return false // fully empty — handled by isEmpty()
+	}
+	if mp.Path == "" && mp.Glob == "" && mp.CwdGlob == "" {
+		return true // vars without a target
+	}
+	if len(mp.Vars) == 0 {
+		return true // target without vars
+	}
+	if mp.CwdGlob != "" && len(mp.Commands) == 0 {
+		return true // cwd_glob without commands
+	}
+	return false
+}
+
+// discardMapping removes the in-progress mapping from the slice and
+// pops the form. Used by the Esc-on-empty fast path and by the
+// Discard branch of the resume/discard prompt.
+func (s *mappingFormScreen) discardMapping() tea.Cmd {
+	if s.idx >= 0 && s.idx < len(s.root.cfg.Mappings) {
+		s.root.cfg.Mappings = append(s.root.cfg.Mappings[:s.idx], s.root.cfg.Mappings[s.idx+1:]...)
+	}
+	return tea.Sequence(emit(popMsg{}), emit(mappingChangedMsg{}))
+}
+
 // maxCursor returns the highest valid cursor row. Cwd mappings get an
 // extra "commands" row above "variables".
 func (s *mappingFormScreen) maxCursor() int {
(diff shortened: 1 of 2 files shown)


Galvill:

PR #170, opened 2026-05-19
fix(tui/mappings): show per-kind hint under the mapping picker (#108)

### LINKED ISSUE #108: TUI: mapping-kind picker has no description for path / glob / cwd — surface a hint line below the menu
## Severity
low (discoverability — new users have to read docs or guess what \"path\" / \"glob\" / \"cwd\" mean when adding their first mapping)

## Where
- `internal/tui/mappings_screen.go:360-361` — kind picker pushes a generic `popupMenuScreen` with `\"path\", \"glob\", \"cwd\", \"Back\"` as the only labels
- `internal/tui/popup_menu.go` — the reusable bordered menu has no slot for per-choice description text

For context, the mapping form's `HelpText()` (`mappings_screen.go:228-231`) already documents the three kinds in prose — but that body is only visible from the form's `?` help, not from the picker itself.

## Problem

When the user picks `< Create New Mapping >` and lands on the kind picker, the modal shows nothing but three bare labels:

```
┌──────────────────┐
│ Mapping kind     │
│                  │
│ ▸ path           │
│   glob           │
│   cwd            │
│   Back           │
└──────────────────┘
```

A new user with no jitenv background has no idea what differentiates them. \"glob\" arguably hints at a pattern, but \"path\" vs \"cwd\" reads as the same thing. \"cwd\" in particular is the only kind that unlocks the wrapper-symlink / interpreter-injection flow (npm / node / pip / python) — that's a big behavioural difference to discover by trial-and-error.

The picker takes up a small corner of the screen; the rest is blank. That's the natural spot to put a one-line explanation of the currently-focused choice.

## Desired

As the cursor moves between `path` / `glob` / `cwd` in the modal, render a hint line just below the modal, in the space that's currently empty. Suggested text (short enough to fit one line at typical terminal widths):

- **path** — One exact file path. The hook intercepts only when the user types that exact path or `./relative`-form of it.
- **glob** — A doublestar pattern over filesystem paths (e.g. `~/scripts/**/*.sh`). Matches multiple files; declaration order resolves conflicts.
- **cwd** — A directory glob. Activates while the shell is inside the matching directory and applies to a configurable list of commands (e.g. `npm`, `node`). Use this for project-scoped env vars that follow a working directory rather than a single script.

Mock-up:

```
┌──────────────────┐
│ Mapping kind     │
│                  │
│ ▸ cwd            │
│   path           │
│   glob           │
│   Back           │
└──────────────────┘

A directory glob. While the shell is inside the matching
directory, env vars are injected when one of the configured
commands runs (npm, node, pip, …). Use for project-scoped
env vars.
```

`Back` would have no hint (or render an empty line so the layout doesn't jump).

## Suggested fix

Two paths:

### 1. Generalise `popupMenuScreen` (preferred for reuse)

Add an optional `descriptions []string` parallel to `choices` on `popupMenuScreen`. When non-empty, render `descriptions[cursor]` below the modal in `View()`:

```go
type popupMenuScreen struct {
    …
    descriptions []string // optional, len matches choices when set
}

// New constructor (keeps the existing one for the no-description case):
func newPopupMenuScreenWithDescs(r *rootModel, heading string,
    fn func(string) tea.Cmd, choices, descs []string) *popupMenuScreen { … }
```

In `View()`:

```go
if len(p.descriptions) == len(p.choices) && p.descriptions[p.cursor] != \"\" {
    b.WriteString(\"\\n    \" + hintStyle.Render(p.descriptions[p.cursor]) + \"\\n\")
}
```

Then update the kind picker call site:

```go
// internal/tui/mappings_screen.go:360-361
return emit(pushMsg{s: newPopupMenuScreenWithDescs(s.root, \"Mapping kind\", cb,
    []string{\"path\", \"glob\", \"cwd\", \"Back\"},
    []string{
        \"One exact file path…\",
        \"Doublestar glob over filesystem paths…\",
        \"Directory glob; activates while shell is in a matching dir…\",
        \"\",
    })})
```

Other popup-menu sites (bag/key/source/mapping context menus) get this for free if a future caller wants to use it; nothing changes for them by default.

### 2. Dedicated kind-picker screen (lower-risk, less reuse)

If touching `popupMenuScreen` feels heavier than warranted, a one-off `kindPickerScreen` that owns its own rendering and keeps the descriptions inline. Same UX, narrower blast radius.

(1) is the cleaner shape because *every* popup picker in the TUI has at least one cryptic-to-newcomers choice that would benefit from the same hint slot. Worth doing once.

## Notes

- Keep descriptions to one line. The popup is bordered with `lipgloss.RoundedBorder()` and currently rendered at the top of an otherwise empty viewport; adding 1-2 lines below is fine, but multi-paragraph blurbs would push the modal off-screen on small terminals.
- The same treatment makes sense for the **\"new source type\"** picker (currently surfaces `local` / `aws` with no explanation) — out of scope for this issue but a natural follow-up once the menu plumbing is in place.
- The form's existing `HelpText()` body (visible via `?`) is the right place for the *long* explanation; the menu-level hint is the *short* one. Don't merge them — they serve different moments.

## Related

- #76 (AWS form duplicated help text) — adjacent area; the schema-level `Help` vs `Placeholder` split proposed there is the form-side equivalent of what this issue asks for at the popup-menu level.

Galvill:
Closes #108. See commit.

### DIFF
--- a/internal/tui/mappings_screen.go
+++ b/internal/tui/mappings_screen.go
@@ -358,7 +358,12 @@ func (s *mappingFormScreen) openKindMenu() tea.Cmd {
 		return emit(popMsg{})
 	}
 	return emit(pushMsg{s: newPopupMenuScreen(s.root,
-		"Mapping kind", cb, "path", "glob", "cwd", "Back")})
+		"Mapping kind", cb, "path", "glob", "cwd", "Back").
+		withHints(map[string]string{
+			"path": "Match one exact filesystem path (e.g. /usr/local/bin/deploy.sh).",
+			"glob": "Match any file under a doublestar pattern (e.g. ~/work/**/*.sh).",
+			"cwd":  "Match when the shell's pwd is under a directory pattern; fires only for listed commands.",
+		})})
 }
 
 func (s *mappingFormScreen) openTargetInput() tea.Cmd {
--- a/internal/tui/popup_menu.go
+++ b/internal/tui/popup_menu.go
@@ -12,16 +12,31 @@ import (
 // "Back" choice is conventional and the screen treats it the same as
 // Esc — the caller can also include a literal "Back" entry to make
 // the affordance visible).
+//
+// Optional per-choice descriptions: callers can populate `hints` (label
+// → hint string) via withHints() to render a one-line description of
+// the currently-focused choice beneath the menu. Used by the
+// mapping-kind picker so first-time users see what path / glob / cwd
+// mean without having to open the form's `?` help (#108).
 type popupMenuScreen struct {
 	root     *rootModel
 	heading  string
 	choices  []string
+	hints    map[string]string
 	cursor   int
 	onChoose func(string) tea.Cmd
 }
 
 func newPopupMenuScreen(r *rootModel, heading string, fn func(string) tea.Cmd, choices ...string) *popupMenuScreen {
 	return &popupMenuScreen{root: r, heading: heading, choices: choices, onChoose: fn}
+}
+
+// withHints attaches a label → description map to the menu. Choices
+// not present in the map render with no hint. Returns the receiver
+// for chaining at construction.
+func (p *popupMenuScreen) withHints(hints map[string]string) *popupMenuScreen {
+	p.hints = hints
+	return p
 }
 
 func (p *popupMenuScreen) Title() string { return "menu" }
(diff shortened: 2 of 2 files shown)


Galvill:

PR #171, opened 2026-05-19
fix(shell/powershell): wire JITENV_HOOK_DEBUG into every decision branch (#107)

### LINKED ISSUE #107: Windows: JITENV_HOOK_DEBUG is largely a no-op in the PowerShell hook (sparse log sites + PS-variable footgun)
## Severity
medium (debuggability gap — when a Windows user can't tell why their mapping isn't firing, they don't have the same observability bash/zsh users do)

## Where
- `internal/shell/snippets/powershell.ps1:74-89` (`__jitenv_chpwd` — logs only inside the catch block)
- `internal/shell/snippets/powershell.ps1:145-176` (`__jitenv_rewrite_buffer` — one log site, several silent early returns)
- `internal/shell/snippets/powershell.ps1:186-198` (AcceptLine handler — no log when it runs, no log when PSReadLine is absent and rewrites are silently disabled)

For comparison: `internal/shell/snippets/bash.sh:68-141` calls `__jitenv_log` at every decision branch (`candidate cmd=…`, `is-mapped rc=…`, `branch=case0/2/*`). Same shape in `zsh.sh`.

## Problem

Setting `JITENV_HOOK_DEBUG=1` in a PowerShell session does not produce the kind of visibility a bash/zsh user gets. Two distinct reasons:

### 1. The PowerShell snippet has almost no debug log sites

Comparing what bash/zsh emit per command (5+ lines covering candidate path, resolution, is-mapped rc, and which branch fired) to what `powershell.ps1` emits today:

| Site | bash/zsh | powershell.ps1 |
|------|----------|----------------|
| Candidate command + resolved path | `__jitenv_log \"candidate cmd=[$cmd] resolved=[$resolved]\"` | Single \"candidate=…\" line on line 166 — only when path resolution succeeded *and* the file exists |
| First token not path-shaped (most common case for `npm` etc.) | early return — no log, same as ps1 | silent (line 161) |
| Resolved file missing | (bash returns silently at the `[[ ! -f ]]` check, also no log) | silent (line 162) |
| is-mapped exit code | `__jitenv_log \"is-mapped rc=$rc\"` | folded into the one \"candidate=… is-mapped=…\" line |
| Which branch ran (0 / 2 / other) | `__jitenv_log \"branch=case0…\"` etc. | no log |
| chpwd activity (per-prompt) | n/a (no chpwd flow on bash/zsh) | only logs from the `catch` block on line 84-86; normal-path activity is invisible at the hook layer |
| PSReadLine absent → AcceptLine handler not registered → path/glob rewrites disabled | n/a | **completely silent** — user gets cwd_glob only, but the hook never tells them why their absolute-path mappings stopped intercepting |

The Go-side `__chpwd` subcommand *does* honour `JITENV_HOOK_DEBUG` and writes `jitenv-chpwd: …` lines to stderr (see `internal/chpwd/chpwd.go:242-251`), so once chpwd actually runs, the user gets reconciliation diagnostics. But every shell-side gate that could short-circuit before reaching the Go side is silent on Windows.

### 2. PowerShell variable vs env var footgun

`$env:JITENV_HOOK_DEBUG = '1'` works.
`$JITENV_HOOK_DEBUG = 1` does **not** — that creates a session-scoped PowerShell variable that the hook never reads (it only checks `$env:JITENV_HOOK_DEBUG`). On bash this distinction doesn't exist: `JITENV_HOOK_DEBUG=1 my-command` is a one-shot export and `export JITENV_HOOK_DEBUG=1` is the persistent form, both visible to subprocesses. A Windows user (especially one switching from bash) is plausibly going to type the wrong form and conclude \"debug doesn't work\".

This is exactly the report behind this issue.

### 1. Add the missing log sites in `powershell.ps1`

Introduce a `__jitenv_log` helper at the top, matching the bash/zsh helper, and call it at the same decision points:

```powershell
function global:__jitenv_log {
    param([string]$msg)
    if ($env:JITENV_HOOK_DEBUG) {
        [Console]::Error.WriteLine(\"jitenv-hook: $msg\")
    }
}
```

Then sprinkle calls into `__jitenv_rewrite_buffer`:

- After the resolve at line 160: `__jitenv_log \"candidate=[$buffer] first=[$first] resolved=[$resolved]\"`
- Before `return $buffer` at line 161: `__jitenv_log \"skip: first token not path-shaped\"`
- Before `return $buffer` at line 162: `__jitenv_log \"skip: resolved path not a file ($resolved)\"`
- After the `$rc` read, log branch=case0/2/* like bash does — keeps the format identical so a user can diff a bash trace against a Windows trace.

And in `__jitenv_chpwd` (not just the catch):

- `__jitenv_log \"chpwd: pid=$PID from='$__JITENV_LAST_PWD' to='$cur'\"` before the `& jitenv __chpwd` call.

And at AcceptLine registration in line 186:

```powershell
} else {
    __jitenv_log \"PSReadLine not present — path/glob interception disabled (cwd_glob still works)\"
}
```

That one line alone closes a major silent-failure mode.

### 2. Make the variable/env-var distinction obvious

Two non-mutually-exclusive options:

a. **At hook load, detect the common mistake.** If `$JITENV_HOOK_DEBUG` (the PS variable, not the env var) is set in the *global* scope and `$env:JITENV_HOOK_DEBUG` is empty, write a one-time hint to stderr:

```powershell
if ((Test-Path Variable:JITENV_HOOK_DEBUG) -and -not $env:JITENV_HOOK_DEBUG) {
    [Console]::Error.WriteLine(\"jitenv-hook: note — set `\$env:JITENV_HOOK_DEBUG=1` (with the env: prefix) to enable hook debug; `\$JITENV_HOOK_DEBUG=1` only sets a session variable.\")
}
```

The hint runs once at hook-load (already guarded by `$global:__JITENV_LOADED`), so it doesn't paper the prompt with a recurring warning.

b. **Document it in `docs/shell-hook.md` (and/or wherever the Windows shell-setup landing page lives).** A one-liner in the troubleshooting section: *\"On PowerShell, debug must be enabled via `\$env:JITENV_HOOK_DEBUG = '1'`. `\$JITENV_HOOK_DEBUG = 1` sets a PowerShell variable, which the hook doesn't read.\"*

(a) is the higher-leverage of the two — most users don't read troubleshooting docs until they're already stuck.

### 3. Verify stderr from `jitenv __chpwd` reaches the user

Line 82: `& jitenv __chpwd \"$PID\" $__JITENV_LAST_PWD $cur | Out-Null`. `Out-Null` consumes stdout, and native-exe stderr in pwsh 7+ normally bypasses the pipeline and is rendered as-is. Worth a manual repro under `$PSNativeCommandUseErrorActionPreference = 'Stop'` (pwsh 7.3+ default in some contexts), which can turn native non-zero exits into thrown exceptions that the try/catch then swallows quietly — masking chpwd's own debug output even when the env var *is* correctly set. If that's actually happening, swap the call shape to capture both streams and re-emit:

```powershell
try {
    & jitenv __chpwd \"$PID\" $__JITENV_LAST_PWD $cur 2>&1 | Out-Null
} catch { … }
```

…or explicitly redirect just the success stream: `> $null`.

## Test plan

- pwsh 7 session, `$env:JITENV_HOOK_DEBUG = '1'`, `cd` into and out of a `cwd_glob` mapped dir → expect `jitenv-hook: chpwd: …` then `jitenv-chpwd: …` lines on stderr.
- Same session, `./mapped-script.exe` → expect `jitenv-hook: candidate=… is-mapped=… branch=case0` and then the script runs through `jitenv run`.
- Same session on a PSReadLine-less image → expect the one-time \"PSReadLine not present\" log on hook load.
- New pwsh session, type `\$JITENV_HOOK_DEBUG = 1` then re-source the hook → expect the one-time \"set \$env:JITENV_HOOK_DEBUG\" hint.

## Related

- The shell-hook design currently treats `JITENV_HOOK_DEBUG` as a contract between the shell snippets and the Go side (`internal/chpwd/chpwd.go:246`). Anywhere that contract lives — agent socket calls, `jitenv run`'s decision tree, the shim — should honour the same env var. A separate issue could audit whether all of them do; the agent side notably doesn't (no grep hits in `internal/agent/`).

Galvill:
Closes #107. See commit.

### DIFF
--- a/internal/shell/snippets/powershell.ps1
+++ b/internal/shell/snippets/powershell.ps1
@@ -30,6 +30,40 @@ $__jitenv_nonceBytes = New-Object byte[] 16
 $env:__JITENV_SESSION_NONCE = -join ($__jitenv_nonceBytes | ForEach-Object { $_.ToString('x2') })
 Remove-Variable -Name '__jitenv_nonceBytes' -Scope Local -ErrorAction SilentlyContinue
 
+# JITENV_HOOK_DEBUG truthiness helper. PowerShell's bare
+# `if ($env:JITENV_HOOK_DEBUG)` treats any non-empty string as true,
+# so the obvious-looking "JITENV_HOOK_DEBUG=0" or "=false" in a
+# Windows env editor silently ENABLE debug — the user's mental model
+# would have those disable it. Normalise to the bash/zsh contract:
+# enabled iff set to something other than the empty string / 0 /
+# false / no / off (case-insensitive). #107
+function global:__jitenv_debug_on {
+    $v = $env:JITENV_HOOK_DEBUG
+    if (-not $v) { return $false }
+    switch ($v.ToLowerInvariant()) {
+        '0'     { return $false }
+        'false' { return $false }
+        'no'    { return $false }
+        'off'   { return $false }
+        ''      { return $false }
+    }
+    return $true
+}
+
+# __jitenv_log mirrors bash.sh / zsh.sh: a single chokepoint that
+# writes a `jitenv-hook: ...` line to stderr when debug is on, and
+# no-ops otherwise. Every decision branch in the hook should call
+# this so a user can see why a command did or did not get rewritten
+# (#107). Writes to stderr via [Console]::Error so it goes through
+# the same channel as native commands rather than PowerShell's
+# error stream (which would render as a red error blob).
+function global:__jitenv_log {
+    param([string]$msg)
+    if (__jitenv_debug_on) {
+        [Console]::Error.WriteLine("jitenv-hook: $msg")
+    }
+}
+
 $global:__JITENV_RUNTIME_DIR = {{RuntimeDir}}
 $global:__JITENV_CFG_PATH    = {{ConfigPath}}
 # Recorded so the shim can tell "this shell typed the command" from
@@ -83,6 +117,7 @@ $global:__JITENV_ORIG_PROMPT = $function:prompt
 
 function global:__jitenv_chpwd {
     $cur = (Get-Location).Path
+    __jitenv_log "chpwd: from=$__JITENV_LAST_PWD to=$cur"
     # No 2>$null on purpose: the chpwd subcommand is silent in normal
     # operation (it only writes to stderr when JITENV_HOOK_DEBUG is
     # set). Swallowing stderr here would hide debug diagnostics and a
@@ -91,9 +126,7 @@ function global:__jitenv_chpwd {
     try {
         & jitenv __chpwd "$PID" $__JITENV_LAST_PWD $cur | Out-Null
     } catch {
-        if ($env:JITENV_HOOK_DEBUG) {
-            Write-Error $_
-        }
+        __jitenv_log "chpwd: error: $_"
     }
     $global:__JITENV_LAST_PWD = $cur
 }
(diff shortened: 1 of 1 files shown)


Galvill:

PR #172, opened 2026-05-19
fix(cli/config,lockfile): refuse a second concurrent TUI session (#166)

### LINKED ISSUE #166: config: prevent concurrent `jitenv config` TUI sessions (lockfile-based) — last-write-wins silently clobbers the other session today
## Severity
medium (data-loss risk — two simultaneous `jitenv config` sessions both read the same on-disk snapshot, edit independently, and the last `Save` silently overwrites the other; no warning, no merge, no detection)

## Where
- `internal/cli/config.go:15-28` — `jitenv config` entrypoint; just calls `tui.Run(configPath)` with no concurrency guard
- `internal/config/edit.go:116-142` — `atomicSave` / `AtomicSave`; atomic on disk but not coordinated with concurrent editors
- `internal/tui/save.go:27` — the save call site; sees no \"someone else also has this open\" signal

## Problem

`jitenv config` (no subcommand) drops into the interactive TUI. It loads the config into memory at start, the user edits, and the eventual save is a tempfile-rename (`config.AtomicSave`). Nothing prevents a second `jitenv config` running in another terminal — and that's the data-loss shape:

1. Terminal A: `jitenv config`. Loads config v1. Adds bag `foo`. Doesn't save yet.
2. Terminal B: `jitenv config`. Loads config v1 (still — A hasn't saved). Adds bag `bar`. Hits Ctrl-S. Wrote config v2 = v1 + bar.
3. Terminal A: hits Ctrl-S. Wrote config v3 = v1 + foo. **bag `bar` is gone.**

No diagnostics fire on either side. A's in-memory view never reflected B's write; B's write was completely discarded.

The atomic-rename in `atomicSave` makes the file itself never half-written, but \"atomic\" is per-write; it doesn't help with overlapping read-modify-write cycles.

### Bonus failure shape: the agent reload

After a save, the TUI pings the agent to reload (`internal/tui/save.go` calls `pingAgentReload`). With concurrent sessions, the agent gets two reload pings for two different snapshots — its final state matches whichever save landed last, even if the user staring at the *first* TUI thinks their save is in effect.

## Desired

A second `jitenv config` invoked while another is running should refuse with a clear error:

```
$ jitenv config
error: another `jitenv config` session is editing the file
  pid:        12345
  started:    2026-05-19T11:08:32Z
  hostname:   workstation
  lockfile:   /home/me/.config/jitenv/.config.toml.lock

Close that session first, or remove the lockfile manually if you're sure the
other process is gone (jitenv config releases the lock automatically on exit
or crash; a leftover lockfile usually means a hung process or remote SSH
session).
```

Read-only entrypoints (`jitenv config show`, `jitenv config validate`) don't need to acquire the lock — they just snapshot the file once. They can optionally *peek* at the lockfile to print a soft note (\"another session is editing; values shown may be stale\") but it's not load-bearing.

### Standard pattern: `flock` on a sibling lockfile

Create `<config-dir>/.config.toml.lock` and `flock(LOCK_EX|LOCK_NB)` it for the lifetime of the TUI session. fd stays open in the process; when the process exits (clean or via SIGKILL), the kernel releases the lock — no stale-lock cleanup needed.

```go
// internal/config/lock.go (new file)

// AcquireEditLock takes an exclusive non-blocking advisory lock on a
// sibling .lock file. Returns a release func and the recorded holder
// info (when the lock is contended). The returned fd stays open until
// release; on process death the kernel releases the lock.
//
//  err = nil          → lock acquired; defer release()
//  err = ErrLocked    → another holder; `info` describes them
//  err = other        → I/O failure
func AcquireEditLock(cfgPath string) (release func(), info LockHolder, err error)

type LockHolder struct {
    PID       int
    Hostname  string
    StartedAt time.Time
}
```

Implementation skeleton:

```go
import \"golang.org/x/sys/unix\" // Linux/macOS
// Windows: golang.org/x/sys/windows, LockFileEx with LOCKFILE_EXCLUSIVE_LOCK|LOCKFILE_FAIL_IMMEDIATELY

func AcquireEditLock(cfgPath string) (func(), LockHolder, error) {
    lockPath := filepath.Join(filepath.Dir(cfgPath), \".\"+filepath.Base(cfgPath)+\".lock\")
    f, err := os.OpenFile(lockPath, os.O_CREATE|os.O_RDWR, 0600)
    if err != nil { … }
    if err := flock(f.Fd(), unix.LOCK_EX|unix.LOCK_NB); err != nil {
        // Read holder info that the current owner wrote on acquisition.
        info := readLockHolder(lockPath) // best-effort; may be empty
        f.Close()
        return nil, info, ErrLocked
    }
    // Write our own holder info for the next would-be-acquirer to read.
    _ = f.Truncate(0)
    _, _ = f.Seek(0, 0)
    _ = json.NewEncoder(f).Encode(LockHolder{PID: os.Getpid(), Hostname: hostname(), StartedAt: time.Now().UTC()})
    return func() { f.Close(); /* lock released by kernel */ }, LockHolder{}, nil
}
```

Use it at `tui.Run`:

```go
// internal/tui/tui.go
release, holder, err := config.AcquireEditLock(cfgPath)
if errors.Is(err, config.ErrLocked) {
    return fmt.Errorf(\"another jitenv config session is editing the file\\n\"+
        \"  pid: %d\\n  started: %s\\n  hostname: %s\\n  lockfile: %s\",
        holder.PID, holder.StartedAt.Format(time.RFC3339),
        holder.Hostname, lockfilePathFor(cfgPath))
}
if err != nil { return err }
defer release()
```

### Why a sibling lockfile and not flock-on-config-toml-directly

- `config.toml` is opened and closed transiently during `Load` / `AtomicSave`. If we held an fd on it for the whole TUI session just for the lock, the rename-replace in `atomicSave` would inherit confusing fd semantics (the new file is a different inode; the lock on the old fd is now meaningless). A separate `.config.toml.lock` keeps the lock-holding fd's lifetime independent of the actual file's inode churn.
- Editors and external tools (`vim ~/.config/jitenv/config.toml`) won't honor an advisory lock anyway — anyone touching the encrypted file directly is opting out of jitenv's lifecycle. The lockfile is strictly for `jitenv config`-vs-`jitenv config` coordination.

### Windows specifics

`LockFileEx(handle, LOCKFILE_EXCLUSIVE_LOCK | LOCKFILE_FAIL_IMMEDIATELY, 0, MAXDWORD, MAXDWORD, &overlapped)` is the analogue. Same lifecycle semantics — the handle's `CloseHandle` releases the lock, and Windows releases it for terminated processes. Same release-on-crash story as Unix.

### Read-only subcommands

`jitenv config show` and `jitenv config validate` don't need to acquire. Optionally `peek_lockfile()` and print a one-line stderr warning when the lockfile holder is alive — keeps users from being surprised when `show` doesn't match a concurrent editor's in-progress state. Cheap, not load-bearing.

### Stale-lockfile recovery

Two layers, in order:

1. The kernel releases the `flock` on process death. The lockfile *body* (PID + timestamp) remains on disk but the lock state itself is gone — the next `jitenv config` acquires successfully because `LOCK_EX|LOCK_NB` only blocks on a live lock.
2. The next acquirer overwrites the lockfile body with its own PID. So even after a crash, no stale-lockfile cleanup script is needed.

The only failure mode where the user has to manually `rm` is the case where the kernel didn't release (theoretical on Unix, can happen if the lockfile lives on certain network filesystems that don't honor advisory locks across nodes). The error message includes the path for exactly that recovery.

## Notes / edge cases

- **`jitenv config init`** creates a fresh file when none exists. It should also take the lock to avoid two concurrent inits clobbering each other on a new install. Same code path; either it's a brand-new lockfile or it bumps into an existing one.
- **Lockfile on tmpfs?** The lockfile lives next to the config (so `~/.config/jitenv/` on Linux, `%APPDATA%\\jitenv\\` on Windows). That's a normal local filesystem; `flock` semantics are well-defined there. The CLAUDE.md call-out about *runtime* directories (XDG_RUNTIME_DIR / TempDir) is unrelated — those are agent paths, not config paths.
- **PID reuse pitfall:** displaying the holder PID is helpful for diagnostics but the PID can be reused. The error message text says \"close that session first, or remove the lockfile manually if you're sure the other process is gone\" — frames the PID as a hint, not a contract.
- **TUI panic during the session:** `defer release()` only runs on clean exit. A runtime panic without recovery would close the fd anyway via the process tearing down — same kernel-cleanup story.
- **`jitenv config` invoked from a script piped to stdin** (`echo ... | jitenv config`): same lock semantics; nothing special. The TUI either runs or refuses.

## Test plan

- Start `jitenv config` in shell A; in shell B, run the same — B exits with the holder-info error.
- Kill -9 shell A's process. Run `jitenv config` again — succeeds (kernel released the lock).
- `jitenv config show` and `jitenv config validate` work in shell B while shell A is editing — no lock contention, optional soft warning printed.
- Unit-test `AcquireEditLock` for the contended/uncontended cases with `t.TempDir()`.
- Cross-platform smoke: run the same test under Linux, macOS, and Windows. The CI matrix already covers all three (per #66).

## Out of scope

- Showing the *other* session a real-time \"someone else just edited the file, here's the diff\" notice. Real-time multi-editor merge is a much bigger feature; this issue is just about *preventing the silent clobber*.
- File-system watchers to detect external edits (vim, sed). Out of scope; the lockfile only coordinates `jitenv config` sessions among themselves.
- Pessimistically locking on every `jitenv hook install` / `jitenv unlock` / agent reload — those are short-lived writes (or pure reads, in `unlock`'s case) that finish before contention realistically matters. If we did want belt-and-braces, the same `AcquireEditLock` is reusable, but the failure mode is dramatically less severe than \"editor vs editor\".

## Related

- The agent's reload-on-save flow (`internal/tui/save.go` → `pingAgentReload`) compounds the concurrent-editor problem by making the *agent* state nondeterministic too. This issue fixes the root cause; the agent reload then behaves predictably.
- #62 (release-please CHANGELOG dedupe) and #67 (squash-merge policy) are unrelated but document the same general principle — single writer per artifact reduces a class of races.

Galvill:
Closes #166. See commit.


Galvill:

PR #173, opened 2026-05-19
fix(tui/mappings): render expanded var count for expand-all refs (#165)

### LINKED ISSUE #165: TUI: mappings list under-counts vars when a mapping uses an "expand whole bag" reference
## Severity
medium (correctness — the count is the *one* signal a user has in the mappings list about what each row will actually inject; \"(1 var)\" for a row that injects ten env vars is actively misleading)

## Where
- `internal/tui/mappings_screen.go:141-142` — render uses `len(mp.Vars)` directly
- `internal/config/config.go:80-86` — `VarRef` with empty `Name` means \"expand the whole bag\" (one ref = N env vars)
- `internal/agent/resolver.go:129-134` — the resolver implements that semantic at fetch time

For supporting context, the bag-key data is already fetched (for local sources, directly from `cfg.Secrets[bag]`; for external sources, asynchronously via `Bagger.Bags(ctx)` in `internal/tui/var_tree_screen.go:214-251`).

## Problem

The mappings list renders each row as:

```go
// internal/tui/mappings_screen.go:140-142
row := fmt.Sprintf(\"%-44s  %s\",
    truncate(mappingLabel(mp), 44),
    dimText(fmt.Sprintf(\"(%d vars)\", len(mp.Vars))))
```

`len(mp.Vars)` counts *references*, not env vars. A `VarRef` with `Name == \"\"` is the \"expand the whole bag/source response\" form (see `config.go:80-86` and `resolver.go:129-134`):

```go
// agent/resolver.go:129-134 — at fetch time, an expand-all ref produces N entries
if v.Name == \"\" {
    for k, val := range raw {
        out[k] = val
    }
    continue
}
```

So a mapping that has a single expand-all reference to a bag with ten keys renders as `(1 var)` in the list but injects ten env vars when the mapping fires. Users who scan the list to confirm \"will this mapping inject `DATABASE_URL`?\" see a small number and miss what's actually there.

### Repro

1. Create a local bag `mybag` with three keys: `A`, `B`, `C`.
2. Add a mapping (any kind) and tick the *bag-level* checkbox in the variables tree (not individual keys).
3. Return to the mappings list. The row shows `(1 var)` — but at runtime three env vars (`A`, `B`, `C`) will be injected.

Same shape for a mapping mixing both: one expand-all ref + two named refs = `(3 vars)` displayed, but actually `3 + len(bag) - 1` injected.

## Desired

Render an accurate count of env vars *the mapping will inject* on a match, not a count of `VarRef` rows. For the example above, `(3 vars)` should read e.g. `(5 vars)` (the three bag keys + the two named refs).

Three sub-cases to handle:

1. **`Name != \"\"`** → contributes 1.
2. **`Name == \"\"` referencing a *local* bag** → contributes `len(cfg.Secrets[v.Ref])`. The TUI already has this data eagerly available — no extra fetch.
3. **`Name == \"\"` referencing an *external* source bag (AWS Secrets Manager, etc.)** → contributes `len(bag.Keys)` once `Bagger.Bags(ctx)` has returned. Until it does, the count is unknown. Two reasonable renderings:
   - `(? vars)` while loading, then refresh once the async result lands.
   - `(N+ vars)` — minimum-known count (number of definitely-injected refs ignoring the unknown expand-all), with a trailing `+` to mark the lower-bound nature.

(a) is preferred — same loading affordance as the var-tree screen already uses. The async fetch is bounded (the var-tree fetch is 15s context timeout) and the result can populate a per-source bag-key cache on `rootModel` that the mapping label reads from.

### 1. Add a helper that counts injected vars

In `internal/tui/mappings_screen.go` (or a new `internal/tui/varcount.go`):

```go
// expandedVarCount returns how many env vars `mp` will inject on a match.
// For each VarRef: a named ref contributes 1; an expand-all ref contributes
// the bag-key count from the source. Returns the count and `known` — false
// when at least one external-source bag-key count is still loading.
func expandedVarCount(cfg *config.Config, bagCache map[string]int, mp *config.Mapping) (count int, known bool) {
    known = true
    for _, v := range mp.Vars {
        if v.Name != \"\" {
            count++
            continue
        }
        // Expand-all reference. Try local source first, then cache.
        if sc, ok := cfg.Sources[v.Source]; ok && sc.Type == \"local\" {
            count += len(cfg.Secrets[v.Ref])
            continue
        }
        if n, ok := bagCache[v.Source+\"|\"+v.Ref]; ok {
            count += n
            continue
        }
        known = false
    }
    return count, known
}
```

`bagCache` lives on `rootModel` and is populated by piggybacking on `loadAsyncBags()` — when the var-tree screen fetches bag listings, store per-bag key counts in the cache so the mappings screen can re-render with accurate numbers next time it's drawn.

### 2. Use it in the render

```go
// internal/tui/mappings_screen.go:140-142 — replace with:
n, known := expandedVarCount(m.root.cfg, m.root.bagKeyCount, &mp)
label := fmt.Sprintf(\"(%d vars)\", n)
if !known {
    label = fmt.Sprintf(\"(%d+ vars)\", n) // or \"(%d vars, loading…)\" — pick one
}
row := fmt.Sprintf(\"%-44s  %s\",
    truncate(mappingLabel(mp), 44),
    dimText(label))
```

`mappingLabel` itself is unchanged.

### 3. Prime the cache eagerly when the mappings screen mounts

The mappings list is one of the first screens a user lands on; if `loadAsyncBags()` is only fired when the var-tree screen opens, the mapping rows would render as `(N+ vars)` until the user manually navigates into the picker. Cheap fix: kick off the same async fetch from the mappings list's `Init()` so the cache is populated by the time the first row is drawn (or shortly after, in which case the screen redraws when `asyncBagsLoadedMsg` arrives).

The cache is per-`rootModel`, so it stays warm across screens for the lifetime of the TUI session. Re-fetching is only needed when sources change — and the existing dirty/save machinery already invalidates downstream caches in adjacent places.

## Notes / edge cases

- **External source whose `Bags()` errors** (e.g. AWS creds wrong, no network). Cache should still be populated with `0` for that source so the display stops showing `(N+)` indefinitely; the var-tree screen surfaces the error separately. The mapping count would then under-report instead of over-mark-as-unknown — but that's not worse than today's status quo for that mapping, and the error is visible elsewhere.
- **Empty bag** (zero keys) selected as expand-all. `len(cfg.Secrets[bag]) == 0` → mapping contributes zero vars from that ref. Today the row says `(1 var)` for the same situation. Either rendering is arguably misleading (a zero-var mapping is itself a bug shape worth flagging — see #163 about partial mappings — but that's a separate fix).
- **Same bag referenced twice in one mapping** (once as expand-all, once as a named key from the same bag): the named key gets shadowed by the expansion at fetch time per `resolver.go:114-141` (later writes win in the map). The count helper should still total both refs because they're both in `mp.Vars`; reporting \"as configured\" is more honest than guessing the runtime overlap.
- The same incorrect counting probably happens anywhere `len(mp.Vars)` is used as a user-facing display number. A grep audit at the same time would be worth it — but `mappings_screen.go:142` is the only hit today.

## Test plan

- Local bag `mybag` with 3 keys, mapping that expands `mybag` only → list shows `(3 vars)`.
- Same mapping plus two named refs to other sources → list shows `(5 vars)`.
- Local bag `empty` with 0 keys, mapping that expands it → list shows `(0 vars)`.
- External AWS source mapping with `Bags()` still loading → list shows `(0+ vars)` initially, then `(N vars)` after the async fetch lands.
- Unit-test `expandedVarCount` directly with stubbed `cfg.Sources` and `bagCache`.

## Related

- #108 (mapping-kind picker descriptions) and #163 (Esc on partial mapping) — same screen, related polish.
- The expand-all semantic is also exposed in the var-tree picker's `bagSel` checkbox (`internal/tui/var_tree_screen.go`). The count rendering there is correct (it shows the unfolded keys). This issue is specifically about the *mappings list* row label.

Galvill:
Closes #165. See commit.

### DIFF
--- a/internal/tui/mappings_screen.go
+++ b/internal/tui/mappings_screen.go
@@ -137,9 +137,19 @@ func (m *mappingsListScreen) View() string {
 	}
 
 	for i, mp := range m.root.cfg.Mappings {
+		count, known := expandedVarCount(m.root.cfg, &mp)
+		suffix := "vars"
+		if !known {
+			// At least one expand-all reference points at a source
+			// whose bag we couldn't size synchronously (typically a
+			// remote source not pre-fetched). Render as a lower
+			// bound so the user sees N+ rather than a misleadingly
+			// precise number.
+			suffix = "vars+"
+		}
 		row := fmt.Sprintf("%-44s  %s",
 			truncate(mappingLabel(mp), 44),
-			dimText(fmt.Sprintf("(%d vars)", len(mp.Vars))))
+			dimText(fmt.Sprintf("(%d %s)", count, suffix)))
 		if i+1 == m.cursor {
 			b.WriteString(" " + labelStyle.Render("▶ ") + listItemFocusedStyle.Render(row) + "\n")
 		} else {
@@ -147,6 +157,40 @@ func (m *mappingsListScreen) View() string {
 		}
 	}
 	return b.String()
+}
+
+// expandedVarCount returns the number of env vars `mp` will inject
+// on a match, expanding "whole bag" references (VarRef with empty
+// Name) by the source's bag size (#165). For local-source expand-all
+// refs the bag size is read directly from cfg.Secrets — no async
+// fetch needed. For remote-source expand-all refs we can't know the
+// size without an HTTP/SDK call we'd rather not run from the
+// mappings-list render path, so the count returns known=false and
+// the caller renders a lower-bound `N+`.
+func expandedVarCount(cfg *config.Config, mp *config.Mapping) (count int, known bool) {
+	if cfg == nil || mp == nil {
+		return 0, true
+	}
+	known = true
+	for _, v := range mp.Vars {
+		if v.Name != "" {
+			// Named ref → exactly one env var.
+			count++
+			continue
+		}
+		// Expand-all reference. Try the synchronous local-source
+		// shortcut first.
+		sc, srcOK := cfg.Sources[v.Source]
+		if srcOK && sc.Type == "local" {
+			count += len(cfg.Secrets[v.Ref])
+			continue
+		}
+		// Remote source or unknown — we'd need an async bag fetch
+		// to size this accurately. Surface the unknown via known=false
+		// and don't add anything to count.
+		known = false
+	}
+	return count, known
 }
 
 func mappingLabel(mp config.Mapping) string {
--- a/internal/tui/mappings_screen_test.go
+++ b/internal/tui/mappings_screen_test.go
@@ -1,0 +1,84 @@
+package tui
+
+import (
+	"testing"
+
+	"github.com/gv/jitenv/internal/config"
+)
+
+// TestExpandedVarCount is the regression for #165: the mappings list
+// rendered `len(mp.Vars)` which counts *references*, not env vars. A
+// VarRef with empty Name expands to N env vars (every key in the bag),
+// so a single expand-all ref to a 10-key bag was shown as "1 var"
+// instead of "10 vars". Local-source expand-alls are counted exactly;
+// remote-source expand-alls return known=false so the renderer shows
+// a lower-bound `N+`.
+func TestExpandedVarCount(t *testing.T) {
+	cfg := &config.Config{
+		Sources: map[string]config.SourceConfig{
+			"localsrc":  {Type: "local"},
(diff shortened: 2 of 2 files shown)


Galvill:

Issue #180, opened 2026-05-25
Add Chocolatey package for Windows installs


Galvill:
## Severity
enhancement (UX — today Windows users have to download `jitenv_X.Y.Z_windows_amd64.zip` from the releases page, extract it manually, add the binary to `%PATH%`, then run `jitenv hook install` to wire PowerShell. macOS users get `brew install Galvill/jitenv/jitenv` after #175; Windows lacks the equivalent one-liner)

## Where
- `.goreleaser.yaml` — new `chocolateys:` block (goreleaser v2 has first-class support; same shape as the existing `homebrew_casks:` we landed in #175)
- `.github/workflows/release.yml` — plumb `CHOCOLATEY_API_KEY` through to the goreleaser step, same `${{ secrets.* }}` pattern already used for `HOMEBREW_TAP_GITHUB_TOKEN` and the macOS notarize secrets
- `packaging/chocolatey/` (new) — supporting files goreleaser doesn't auto-generate:
  - `tools/chocolateyInstall.ps1` — placement script (drops `jitenv.exe` into `$toolsDir`, lets choco's shim machinery put it on `%PATH%`)
  - `tools/chocolateyUninstall.ps1` — counterpart removal
  - `tools/LICENSE.txt` — copy of repo LICENSE (chocolatey.org requires it inside the package)
  - `tools/VERIFICATION.txt` — points at the upstream `.zip` URL + SHA256, so moderators can confirm the package matches an official release
- `README.md` — new "Install with Chocolatey" subsection alongside the Homebrew one

## Problem

`jitenv` already cross-compiles for Windows (`amd64` + `arm64`) and ships zips on every release. The hook + agent + named-pipe transport are tested on Windows. What's missing is the bootstrap step a typical Windows user expects:

```
choco install jitenv
```

Without that:
- New users hit a `.zip → extract → put on PATH` flow that is friction at exactly the moment we're trying to win them.
- Upgrades are manual (download new zip, replace old binary). `choco upgrade jitenv` fixes that.
- SmartScreen still warns on first run (binary isn't Authenticode-signed yet — tracked in #39), but the choco package can document the unblock workaround in its description, same way the Homebrew cask currently documents the macOS `xattr -d com.apple.quarantine` workaround until #13's notarization lands.

## Desired flow

```
PS> choco install jitenv
Chocolatey v2.x.x
Installing the following packages: jitenv
...
jitenv v0.8.0 installed.

PS> jitenv hook install
[...]

PR #183, opened 2026-05-25
fix: stop injecting twice in turbo workers + stop terminal-query leak (#182)

### LINKED ISSUE #182: Double env-injection in turbo workers + terminal-query leak on every jitenv invocation
## Summary

Two interacting bugs surface when running `jitenv` under a project that uses turbo (or any sandboxer that filters child env) plus a real terminal. Both reproduce on the current main; neither was introduced by recent PRs.

## Reproducer

```sh
cd ~/my-monorepo                 # has a cwd_glob mapping with commands = ["npm"]
npm run dev                      # → sh scripts/run-dev.sh → turbo --parallel
```

The `jitenv: injected N variables` notice appears once for the top-level `npm run dev` (correct), then a SECOND and THIRD time, one per turbo worker (incorrect):

```
jitenv: injected 8 variables                       # ← top-level — fine
...
@nextvill/parent-frontend:dev: jitenv: injected 8 variables    # ← duplicate
^[]11;rgb:0000/0000/0000^[\^[[30;32R@nextvill/web:dev: jitenv: injected 8 variables    # ← duplicate
^[]11;rgb:0000/0000/0000^[\^[[30;56R@nextvill/web:dev:
```

The garbled `^[]11;…^[\^[[30;32R` chunks are the terminal's *responses* to OSC 11 (background color query) + CPR (`\e[6n` cursor-position query). They're being emitted by jitenv on every invocation and leaked into turbo's output stream.

### Mechanism

`internal/shim/shim.go` short-circuits re-injection via two env markers (security #120):
- `__JITENV_SESSION_NONCE` — minted by the bash/zsh hook
- `__JITENV_INJECTED` — stamped by the first shim hop in an execve chain, value = the nonce

`shim.injectionAlreadyApplied()` returns true only when both are present and equal (constant-time compare). On match, the shim skips fetch, skips runnotice, and just execs.

When `__JITENV_INJECTED` and/or `__JITENV_SESSION_NONCE` get *stripped* between the top-level shell and a downstream worker, the shim re-injects from scratch and re-prints the notice.

### Root cause

**Turbo 2.x defaults to "Strict Environment Mode"** (https://turborepo.com/docs/crafting-your-repository/using-environment-variables#strict-mode). Children only receive env vars that turbo explicitly allows via `globalEnv`, `globalPassThroughEnv`, `env`, or `passThroughEnv`. Anything outside those lists is stripped — including the two `__JITENV_*` markers.

Reproducible by any env-filtering child mechanism: firejail / bwrap / kubernetes Pod env / some IDE-embedded shells.

### Possible fixes

1. **Document + add a recipe to README.** Tell turbo users to add `globalPassThroughEnv: ["__JITENV_INJECTED", "__JITENV_SESSION_NONCE", "__JITENV_SHELL_PID", "__JITENV_WRAP_DIR"]` to `turbo.json`. Lowest-effort.
2. **Detect already-injected via process tree ancestry instead of env.** Walk `getppid()` chain looking for a process whose `/proc/<pid>/environ` contains `__JITENV_INJECTED`. Survives env stripping. Linux-only, more brittle, but no user setup.
3. **Write a marker file under `XDG_RUNTIME_DIR/jitenv/shells/<shell-pid>/injected-<chain-id>`** that survives env stripping. A subsequent shim entry under the same shell-pid sees the file and short-circuits. Cross-platform, needs careful cleanup.

Recommendation: ship #1 immediately (one-paragraph README addition), keep #2/#3 as a real follow-up.

### Out-of-scope discovery: `__JITENV_SHELL_PID` is also stripped

`shim.shouldInject()` returns true when `__JITENV_SHELL_PID` is empty (the "no hook loaded" fallback path). In a strict-env turbo worker, that var is gone → `shouldInject` returns true → fetch fires. This is what actually re-triggers the path. Worth noting in the fix design.

### Mechanism

Bubble Tea v1.3.10 ships this at `vendor/github.com/charmbracelet/bubbletea/tea_init.go`:

```go
func init() {
    // XXX: This is a workaround to make assure that Lip Gloss and Termenv
    // query the terminal before any Bubble Tea Program runs...
    _ = lipgloss.HasDarkBackground()
}
```

`lipgloss.HasDarkBackground()` calls `termenv.Output.backgroundColor()`, which `termenv_unix.go:269,272` writes to the tty:

```go
fmt.Fprintf(tty, OSC+"%d;?"+ST, 11)   // ]11;?
fmt.Fprintf(tty, CSI+"6n")            // [6n
```

This `init()` runs at *program startup* — before main, before Cobra dispatch, before we know what subcommand the user typed. It fires on every jitenv invocation, including `jitenv version`, `jitenv __shim`, `jitenv run`, etc.

I confirmed by hex-dumping captured stdout of `jitenv version`:

```
00000060: 6e20 7465 726d 696e 616c 3e5d 0a1b 5d31  n terminal>]..\e]1
00000070: 313b 3f1b 5c1b 5b36 6e6a 6974 656e 7620  1;?\e\\\e[6njitenv
```

`\e]11;?\e\\` + `\e[6n` written before any jitenv output.

### Why it's only now visible

In an interactive shell, the shell consumes terminal escape responses on stdin. In turbo's child-pty model, the worker's stdin isn't the user's terminal — turbo's stdout-capture pipeline picks up the response bytes and emits them as worker output, prefixed with the task label.

The bug has been latent since whenever the TUI was added; turbo just exposes it.

### Import chain that pulls bubbletea in

```
cmd/jitenv/main.go
└── internal/cli (subcommands aggregator)
    ├── internal/cli/config.go   →  imports internal/tui  (for `jitenv config`)
    └── internal/cli/clone.go    →  imports internal/tui  (#181, post-clone handoff)

internal/tui
└── github.com/charmbracelet/bubbletea  ← init() runs at program startup
```

The transitive import means every jitenv binary invocation pays the OSC-query cost.

### Possible fixes

1. **Split `jitenv config` (and the post-clone TUI handoff) into a separate binary** `jitenv-tui`. `cli/config.go` execs it; `cli/clone.go` execs it. The main `jitenv` binary's import graph then excludes bubbletea/lipgloss/termenv. Cleanest semantically; goreleaser would need to build + distribute two binaries.
2. **Upgrade Bubble Tea to v2** — the v1.3.10 init comment says "This workaround will be removed in v2." API is breaking; non-trivial migration.
3. **Vendor a stripped bubbletea** that removes `tea_init.go`. Brittle, hard to maintain across upstream changes.
4. **Set `TERM=dumb` (or unset COLORTERM) in main()'s very first line.** Doesn't help — `init()` already ran before main().

Recommendation: tracking issue → #1 (binary split) is the right long-term fix; #2 follows whenever bubbletea v2 lands.

## Test plan once fixed

- `xxd <(jitenv version 2>&1)` shows no OSC 11 / CPR bytes.
- `npm run dev` in a turbo monorepo with `mapped cwd_glob` shows `jitenv: injected N variables` exactly once, regardless of how many workers turbo spawns.
- Adding `globalPassThroughEnv: ["__JITENV_INJECTED", "__JITENV_SESSION_NONCE", "__JITENV_SHELL_PID", "__JITENV_WRAP_DIR"]` to a turbo.json restores correct behaviour (sanity check that Bug A's mechanism analysis is right).

## Related

- security #120 — session-nonce mechanism that this issue exercises.
- #77 — original double-injection fix; this is a sibling failure mode under env stripping.
- #71 — same propagation pattern for the agent-down `__JITENV_AGENT_WARNED` marker; will need the same fix.

Galvill:
Closes #182.

## Summary

Two commits, two bugs:

### Bug A — `fix(shim): on-disk marker file as env-stripping fallback`

Turbo 2.x's Strict Environment Mode (and firejail / bwrap / sandboxer variants) strips env vars before spawning children, which defeats the existing `__JITENV_INJECTED` / `__JITENV_SESSION_NONCE` bypass (security #120). The shim then re-fetches and re-prints `jitenv: injected N variables` per worker.

Fix: add an on-disk marker file at `<wrap-dir>/../injected` (i.e. `<runtime>/shells/<shell-pid>/injected`) as a second bypass channel. The wrap-dir path survives env stripping because PATH itself has to be preserved for the wrapper to be found in the first place. File presence — not contents — gates the bypass; the nonce is still written so future tools can do per-chain matching. Cleaned up by `agent.GcOrphanShells` when the shell exits.

### Bug B — `feat(cli): split TUI into separate jitenv-tui binary`

Bubble Tea v1.3.10's `tea_init.go` runs `lipgloss.HasDarkBackground()` at package init, sending OSC 11 + CPR queries to the tty. In an interactive shell the responses are consumed by the shell; in turbo's pty (and any other context that doesn't manage tty responses) the bytes leak into captured output. Because `internal/cli` previously imported `internal/tui`, bubbletea was linked into the main binary and its init fired on EVERY jitenv invocation.

Fix: factor the TUI into a new `cmd/jitenv-tui` binary. `internal/cli/config.go` re-execs it via `os/exec` when the user runs `jitenv config`. The main binary's import graph no longer contains bubbletea / lipgloss / termenv.

## Verified

```
$ xxd <(jitenv version 2>&1) | grep -cE "1b 5d|1b 5b"
0
$ xxd <(jitenv-tui 2>&1) | grep -cE "1b 5d|1b 5b"
1   # expected — TUI is allowed to query because the user is in it
```

End-to-end in the e2e debian container with a fresh deb install:
- Both binaries land at `/usr/bin/jitenv` and `/usr/bin/jitenv-tui`.
- `jitenv version` byte-for-byte clean (no escape responses).
- `jitenv config` resolves the sibling binary, prompts for the passphrase, and renders the TUI.
- New `TestShimSuppressesInjectionViaMarkerFile` exercises the env-stripping branch directly.

Full test suite + `golangci-lint v1.62.2` green.

## Packaging changes

| Artifact | Before | After |
|---|---|---|
| `dist/jitenv_<v>_<os>_<arch>.tar.gz` | jitenv | jitenv + jitenv-tui |
| deb / rpm | `/usr/bin/jitenv` | `/usr/bin/jitenv` + `/usr/bin/jitenv-tui` |
| Homebrew cask | `binary "jitenv"` | `binaries: [jitenv, jitenv-tui]`, both un-quarantined post-flight |
| Makefile `build` / `install` | jitenv only | jitenv + jitenv-tui |

The forthcoming Chocolatey package (#180) will need the same treatment when it lands.

## Binary resolution chain (in jitenv → jitenv-tui handoff)

Most-specific first; the first hit wins:

1. `JITENV_TUI_BIN` env var (test / dev override).
2. Sibling to the running jitenv executable — typical packaging case (both binaries in the same `bin/` dir).
3. `PATH` lookup.

A clear error message points at the install if neither resolves, in case a user upgrades manually and forgets the second binary.

## Test plan

- [x] `go test ./...` green.
- [x] `golangci-lint v1.62.2 run ./...` clean.
- [x] `xxd <(jitenv version)` produces zero OSC 11 / CPR bytes.
- [x] e2e debian: deb install, both binaries on disk, TUI launches via re-exec.
- [x] Marker-file test covers the env-stripping bypass directly.
- [ ] Manual: a real turbo monorepo (the user's reproducer) prints `jitenv: injected N variables` exactly once across all workers — to be verified by the issue author after merge.

## Out-of-scope (deferred)

- Process-tree ancestry detection (the alternative to file-marker proposed in #182). The file-marker is cross-platform; process-tree walking is Linux-only and more brittle. We can revisit if the file-marker proves insufficient.
- Bubble Tea v2 upgrade — separate effort, breaking API changes.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -27,6 +27,34 @@ builds:
       - -X github.com/gv/jitenv/internal/version.Commit={{.ShortCommit}}
       - -X github.com/gv/jitenv/internal/version.Date={{.Date}}
 
+  # The interactive TUI lives in a separate binary so the main
+  # jitenv binary's import graph stays free of Bubble Tea / Lip
+  # Gloss / termenv. Those libs send OSC 11 (background color) +
+  # CPR (\e[6n) queries to the tty at package-init time; in
+  # contexts where the parent doesn't consume the responses
+  # (turbo strict-env pty being the canonical example) the
+  # responses leak into the user's output stream. See #182 bug B.
+  # `jitenv config` re-execs this binary via internal/cli/config.go.
+  - id: jitenv-tui
+    main: ./cmd/jitenv-tui
+    binary: jitenv-tui
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
+      - windows
+    goarch:
+      - amd64
+      - arm64
+    flags:
+      - -trimpath
+    ldflags:
+      - -s -w
+      - -X github.com/gv/jitenv/internal/version.Version={{.Version}}
+      - -X github.com/gv/jitenv/internal/version.Commit={{.ShortCommit}}
+      - -X github.com/gv/jitenv/internal/version.Date={{.Date}}
+
 archives:
   - id: jitenv
     # Unix archives are tar.gz; Windows archives are .zip so users can
@@ -36,7 +64,13 @@ archives:
     format_overrides:
       - goos: windows
         formats: [zip]
-    name_template: "{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
+    # Both binaries ship in the same archive so a brew/dpkg/manual
+    # install gets the pair atomically. `ids:` is the v2-spelling
+    # of the build allowlist.
+    ids:
+      - jitenv
+      - jitenv-tui
+    name_template: "jitenv_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
     files:
       - LICENSE
       - README.md
@@ -50,6 +84,9 @@ archives:
 nfpms:
   - id: jitenv
     package_name: jitenv
+    ids:
+      - jitenv
+      - jitenv-tui
     vendor: Galvill
     homepage: https://github.com/Galvill/jitenv
     maintainer: Galvill <nexvill.ai@gmail.com>
@@ -106,6 +143,13 @@ homebrew_casks:
     homepage: https://github.com/Galvill/jitenv
     description: Just-in-time env-var injection from pluggable secret sources, scoped to per-command process trees
     skip_upload: auto
+    # Ship both halves of the binary split (#182 bug B): jitenv runs
+    # everything except the TUI; jitenv-tui is the re-exec target for
+    # `jitenv config`. The cask links both into the Homebrew prefix
+    # bin/ so `which jitenv-tui` resolves.
+    binaries:
+      - jitenv
+      - jitenv-tui
     commit_author:
       name: goreleaserbot
       email: bot@goreleaser.com
@@ -120,6 +164,8 @@ homebrew_casks:
           if system_command("/usr/bin/xattr", args: ["-h"]).exit_status == 0
             system_command "/usr/bin/xattr",
                            args: ["-dr", "com.apple.quarantine", "#{staged_path}/jitenv"]
+            system_command "/usr/bin/xattr",
+                           args: ["-dr", "com.apple.quarantine", "#{staged_path}/jitenv-tui"]
           end
 
 checksum:
--- a/Makefile
+++ b/Makefile
@@ -17,17 +17,19 @@ LDFLAGS := -s -w \
 
 build:
 	$(GO) build -trimpath -ldflags "$(LDFLAGS)" -o bin/$(BIN) ./cmd/jitenv
+	$(GO) build -trimpath -ldflags "$(LDFLAGS)" -o bin/$(BIN)-tui ./cmd/jitenv-tui
 
-# Cross-compile the Windows binary. Stage 1 (#80) made the codebase
-# compile for GOOS=windows; the runtime port lands in #86-91. Useful
-# for reviewers verifying a Windows-touching PR really produces an
-# executable — `go build ./...` is a compile-check only and does not
-# emit a .exe.
+# Cross-compile the Windows binaries. The TUI ships as a separate
+# binary (#182 bug B): the main jitenv binary stays free of
+# Bubble Tea / Lip Gloss imports so it doesn't query the terminal
+# on every invocation.
 build-windows:
 	GOOS=windows GOARCH=amd64 $(GO) build -trimpath -ldflags "$(LDFLAGS)" -o bin/$(BIN).exe ./cmd/jitenv
+	GOOS=windows GOARCH=amd64 $(GO) build -trimpath -ldflags "$(LDFLAGS)" -o bin/$(BIN)-tui.exe ./cmd/jitenv-tui
 
 install: build
 	install -Dm0755 bin/$(BIN) $(PREFIX)/bin/$(BIN)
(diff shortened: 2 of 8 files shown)


Galvill:

PR #184, opened 2026-05-25
fix(shim): recover wrap dir from PATH when env vars are stripped (#182)

Galvill:
Stacked on #183.

## Symptom

User reported that the marker-file fix from #183 worked at the top level (no `jitenv: injected N variables` for the outer `npm run dev`) but the turbo workers still emitted the notice twice:

```
gv@BEAST:~/Code/NexVill-Project/NexVill (main)$ npm run dev
...
@nextvill/parent-frontend:dev: jitenv: injected 8 variables
@nextvill/web:dev: jitenv: injected 8 variables
```

## Root cause

Turbo strict env strips `__JITENV_WRAP_DIR` *as well as* `__JITENV_INJECTED` / `__JITENV_SESSION_NONCE`. Combined with the worker's `argv[0] = "npm"` (execvp-by-name semantics), the shim's fallback `filepath.Dir(argv[0])` produces `"."`. That value then breaks:

1. **PATH exclusion** in `lookPathExcluding(".", invokedAs)` — the wrap dir doesn't equal `"."`, so it's not excluded, so the wrapper is what the lookup returns. Infinite shim re-exec in theory.
2. **Marker-file check** — `shellDirFromWrap(".")` returns `""` because `Base(".")` ≠ `"bin"`, so `markerFileSays(".")` is false, so the bypass doesn't fire, so the worker re-injects and re-prints the notice.

## Fix

When neither `__JITENV_WRAP_DIR` nor `filepath.Dir(argv[0])` produces a wrap-dir-shaped path, scan `$PATH` for an entry that:

- lives **strictly** under the agent's per-user runtime dir (the `HasPrefix(abs+sep, shellsAbs+sep)` + `abs != shellsAbs` check rejects sibling-dir downgrades like `<runtime>/shells-evil`),
- matches the shape `<…>/<positive-int>/bin`, AND
- actually exists on disk.

The runtime dir is already 0700 owned by the user (security #117 verifies this at agent startup), so a recovered PATH entry rooted there is no weaker to trust than the agent's socket from the same dir.

## Test

New `TestShimRecoversWrapDirFromPath` builds the full turbo-worker reproducer:

- wrap dir under `<XDG_RUNTIME_DIR>/jitenv/shells/<pid>/bin`,
- marker file in `<shellDir>/injected`,
- env stripped of all `__JITENV_*` vars,
- `argv[0]="fakecmd"` (basename, mirroring how `execvp("npm", argv)` leaves it).

The shim must short-circuit via the recovered marker. Passes.

## Test plan

- [x] New + existing shim tests green (`go test ./internal/shim/`).
- [x] Full repo test suite green (`go test ./...`).
- [x] `golangci-lint v1.62.2` clean.
- [ ] Manual: the issue author re-runs `npm run dev` and confirms exactly one notice (or zero, if a previous run in the same shell already dropped the marker file).

### DIFF
--- a/internal/shim/shim.go
+++ b/internal/shim/shim.go
@@ -92,6 +92,24 @@ func run(invokedAs string, args []string) error {
 	if selfDir == "" {
 		selfDir = filepath.Dir(firstArg())
 	}
+	// Env-stripping recovery (#182): when an intermediate process
+	// (turbo strict env mode is the canonical case; firejail / bwrap
+	// / sandboxer variants behave the same) drops __JITENV_WRAP_DIR
+	// before exec, and argv[0] is the bare basename ("npm") because
+	// the caller used execvp-by-name, both lookups above land at "."
+	// — which then breaks both the PATH-exclusion (infinite re-exec
+	// loop) AND the marker-file check (the double-injection symptom
+	// in turbo workers). Recover by scanning PATH for an entry that
+	// lives under the agent's per-user runtime dir and matches the
+	// <runtime>/shells/<pid>/bin shape. The agent verifies that
+	// runtime dir is 0700 owned by the current user (security #117),
+	// so trusting a PATH entry rooted there is no weaker than
+	// reading the agent's socket from the same dir.
+	if !isWrapDirShape(selfDir) {
+		if recovered := findWrapDirInPath(); recovered != "" {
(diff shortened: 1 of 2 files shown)


Galvill:

PR #185, opened 2026-05-25
fix(shim): scope injection marker to one command chain via pgid (#182)

Galvill:
Stacked on #184. Closes the user-reported follow-up on #182.

## Symptom

After PRs #183/#184 fixed the per-worker double-inject, the user reported a new regression:

> after one time the injected file marker is not cleaned and no more injections happen.

Sequence:
1. `npm run dev` → first shim writes marker file → workers bypass cleanly → notice fires exactly once. ✓
2. Ctrl+C the tree.
3. `npm run dev` again → marker file still on disk → first shim short-circuits → **no notice, no env injection**, downstream code crashes on missing `SESSION_SECRET`.

## Root cause

The marker's lifetime was "per shell session" (cleaned up by `agent.GcOrphanShells` only on shell exit). Correct lifetime is "per command chain". The mismatch let stale markers suppress unrelated follow-on commands.

## Fix

Scope the marker to the chain's process-group id (pgid). Write the caller's pgid into the file at first-inject; bypass only fires when the reader's own pgid matches the file content.

**Why pgid works:**

- Bash/zsh job-control puts each foreground command in its own process group via `setpgid(2)`. All descendants of that command — `npm → sh → turbo → workers → re-wrapped npm` — share the pgid unless they explicitly `setpgrp` (well-behaved tools don't). Workers in the same chain see "match" → bypass cleanly.
- A different command in the same shell (the user's next `npm run dev`, or a `git push` between two npms) is a fresh job with a fresh pgid. Reading the stale marker yields a "mismatch" → fresh inject → notice fires.

**Cleanup**: nothing extra needed. Stale markers from finished chains just mismatch, get overwritten on the next inject, and are removed with the shell dir by `agent.GcOrphanShells` at shell exit.

## Tests

- New `TestShimMarkerFromForeignPgidDoesNotBypass` reproduces the user's failure mode directly: drop a marker with a wrong pgid (1, init), confirm the shim does NOT bypass (warning fires, real binary runs).
- Updated `TestShimSuppressesInjectionViaMarkerFile` and `TestShimRecoversWrapDirFromPath` to write the test's own pgid into the marker so they still exercise the bypass path under the new contract.
- Full repo `go test ./...` + `golangci-lint v1.62.2` clean.

## Platform

`pgid` is a Unix concept. `currentPgid()` returns 0 on Windows so `markerFileSays`/`writeMarkerFile` silently no-op there; the env-marker channel still works under any Windows context that doesn't strip env (which is the common case — Windows has no real strict-env analog at the shell level).

### DIFF
--- a/internal/shim/pgid_unix.go
+++ b/internal/shim/pgid_unix.go
@@ -1,0 +1,23 @@
+//go:build !windows
+
+package shim
+
+import "syscall"
+
+// currentPgid returns the calling process's process-group id. Used
+// by the marker-file machinery (#182) to identify a command chain:
+// children of the first shim hop inherit its pgid through fork +
+// execve, so they all see the same value, while a new command from
+// the typing shell gets a fresh pgid via bash/zsh job-control
+// setpgid(2) — exactly the granularity we want.
+//
+// Returns 0 on any error; callers treat 0 as "no chain identity" so
+// the marker bypass simply doesn't fire and the shim falls back to
+// fresh injection.
+func currentPgid() int {
+	pgid, err := syscall.Getpgid(syscall.Getpid())
+	if err != nil {
+		return 0
+	}
+	return pgid
+}
--- a/internal/shim/pgid_windows.go
+++ b/internal/shim/pgid_windows.go
@@ -1,0 +1,11 @@
+//go:build windows
+
+package shim
+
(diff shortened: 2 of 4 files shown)


Galvill:

PR #186, opened 2026-05-26
fix(shim,chpwd): scope injection marker via PROMPT_COMMAND cleanup (#182)

Galvill:
Stacked on #185. Closes the user-reported follow-up to that PR.

## Symptom

After #185's pgid-scoping fix, the user re-tested with `npm run preview` (`turbo`-based command not previously exercised) and double-injection came back:

```
gv@BEAST:~$ npm run preview
jitenv: injected 8 variables                       # ← top-level, fine
...
@nextvill/parent-frontend:preview: jitenv: injected 8 variables    # ← duplicate
@nextvill/web:preview: jitenv: injected 8 variables                # ← duplicate
```

## Root cause

Turbo puts each worker in its own process group — probably for signal isolation, so a SIGINT to one worker doesn't propagate to siblings. So the worker's pgid is NOT the top-level shim's pgid, the pgid-match in `markerFileSays` returns false, the worker re-injects, the notice fires per worker.

**pgid was the wrong granularity.** Too narrow. Session-id would be too broad (one session = one shell = many commands). The right scope is exactly what bash/zsh themselves call "one command": **between two `PROMPT_COMMAND` fires**. The shell hook already fires `jitenv __chpwd` there for cwd_glob wrapper reconciliation — piggy-back on it.

## Fix

1. **`internal/shim/shim.go`** — `markerFileSays` reverts to file-presence (no content match).
2. **`internal/chpwd/chpwd.go`** — `Run` unlinks `<shellsDir>/<pid>/injected` at the top, BEFORE the unchanged-state short-circuit. The cleanup must run even when pwd + cfg mtime didn't change (which is the common `npm run dev` foreground case).

## Lifetime semantics

| When | Marker | Bypass? |
|---|---|---|
| First shim hop of a command | written | n/a (first hop fetches) |
| Worker / chained interpreter, same command | present | yes — bypass |
| `PROMPT_COMMAND` fires (command done) | unlinked by chpwd | n/a |
| Next user command's first hop | absent | n/a (fresh inject) |

## Tests

- New `TestRunUnlinksInjectionMarker` in `internal/chpwd/chpwd_test.go` drops a marker by hand, calls `Run` with same-pwd same-mtime (short-circuit branch), confirms the marker is gone afterward.
- Dropped `TestShimMarkerFromForeignPgidDoesNotBypass` (#185's regression test) — the foreign-pgid distinction no longer exists; file-presence is the contract.
- Existing marker-bypass tests now write `"any-content"` into the file since content is informational only.
- Full repo `go test ./...` + `golangci-lint v1.62.2` clean.

## What about non-interactive contexts?

Scripts / CI jobs without the shell hook → no `PROMPT_COMMAND` → no chpwd cleanup. But:

- The marker dir lives under `<runtime>/jitenv/shells/<pid>/`, and `agent.GcOrphanShells` removes the whole dir when the shell exits.
- CI jobs typically run one command per shell anyway, so the cross-command suppression bug doesn't apply.

## Code that survived the iteration

`internal/shim/pgid_unix.go` + `pgid_windows.go` stay — `writeMarkerFile` still records the pgid as the file's content for ad-hoc debugging ("which pgrp dropped this?"). Not load-bearing; harmless if removed in a future cleanup.

### DIFF
--- a/internal/chpwd/chpwd.go
+++ b/internal/chpwd/chpwd.go
@@ -69,6 +69,28 @@ func Run(args []string) error {
 	wrapDir := paths.ShellWrapDir(pid)
 	mtimePath := lastMtimePath(paths, pid)
 
+	// Scope the shim's on-disk injection marker (#182) to "one user
+	// command" by unlinking it on every prompt fire. The marker
+	// lives at <shellsDir>/<pid>/injected; the shim creates it after
+	// a successful first-hop inject so descendants (turbo workers,
+	// chained interpreters) bypass instead of double-fetching. Once
+	// the user's command tree has exited and bash/zsh fires
+	// PROMPT_COMMAND → __chpwd, we delete the file so the NEXT
+	// command starts clean and gets a fresh injection + notice.
+	//
+	// Best-effort: a stat/unlink failure (file missing — first
+	// prompt of the session, or already cleaned up) is fine. Run
+	// BEFORE the short-circuit-on-unchanged-state below: the marker
+	// must be cleaned even when pwd + cfg mtime didn't change, which
+	// is the common case for a foreground `npm run dev` that
+	// completes in the same dir.
+	markerPath := filepath.Join(filepath.Dir(wrapDir), "injected")
+	if err := os.Remove(markerPath); err == nil {
+		debugLog("removed injection marker %s", markerPath)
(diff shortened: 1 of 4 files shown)


Galvill:

Issue #203, opened 2026-05-27
cwd_glob: wrapper reconcile lags one prompt — external edit at an idle prompt isn't seen until the next prompt fires


Galvill:
## Summary

For `cwd_glob` mappings, the per-shell wrapper symlinks are only reconciled by `__chpwd`, which the hooks drive from `PROMPT_COMMAND` (bash) / `precmd` (zsh) — i.e. just before a prompt is drawn. If a `cwd_glob` mapping is edited from **another process** (another terminal, or a hand-edit) while a shell sits at an idle prompt, that shell's wrapper dir is stale until its **next** prompt fire. The first command after the edit runs against the pre-edit wrapper state.

Discovered while investigating the cwd-mapping staleness issues (#201). Third of three staleness layers; the most benign and the most inherent to the prompt-driven model.

## Severity

Low. Self-heals on the next prompt (every command is followed by a `PROMPT_COMMAND` → `__chpwd`). It only bites the single command typed at an idle prompt whose state predates an edit made by a *different* process. Same-shell edits via the TUI are fine: the post-TUI prompt fire reconciles. Now that #201 makes `__chpwd` clear the command hash on change, the next prompt fully recovers.

## Evidence

- `internal/shell/snippets/bash.sh`: `PROMPT_COMMAND="__jitenv_chpwd${PROMPT_COMMAND:+;$PROMPT_COMMAND}"` and a load-time call. zsh: `precmd_functions+=(__jitenv_chpwd)`.
- `internal/chpwd/chpwd.go` `Run` short-circuits on `oldPwd == newPwd && curMtime == lastMtime`; it reconciles on a pwd change or a config mtime change — but only when invoked, i.e. on a prompt fire.
- `TestCwdGlob_ExternalEditStaleUntilReconcile` (added in #201) demonstrates: with the command hash factored out, an edit that newly maps the current dir is not picked up until `__jitenv_chpwd` runs again.

## Options

1. **Accept + document.** It self-heals on the next prompt; the cost of closing it fully is likely not worth it. Document the one-prompt lag for multi-terminal workflows.
2. **Reconcile opportunistically in the DEBUG trap / accept-line widget.** The bash DEBUG trap and zsh `accept-line` widget already run per command; they could trigger a reconcile for bare commands before exec. Downside: adds a config stat (and possibly a fork) to the hot path of every bare command, and bare commands are exactly what `cwd_glob` targets.
[...]

PR #204, opened 2026-05-27
fix(agent): auto-reload config when it changes on disk (#202)

### LINKED ISSUE #202: agent: in-memory config goes stale on a hand-edit of config.toml (no reload / no file-watch)
## Summary

A running agent only reloads its config on an explicit reload op — there is no user-facing `jitenv reload` command and no file-watch. So a **hand-edited** `config.toml` (or any edit made by a tool other than the TUI / `clone`) is never seen by a running agent: it keeps serving env values from its stale in-memory config until the next `lock`/`unlock` or a TUI save.

Discovered while investigating the cwd-mapping staleness issues (#201). This is one of the three staleness layers; #201 fixes the shell-command-hash layer, this issue tracks the agent layer.

## Severity

Medium. For TUI users it's a non-issue (the TUI pings reload after every save — `internal/tui/tui.go` `pingAgentReload`). It bites users who edit `config.toml` directly or with external tooling, and it can be a correctness/security concern: combined with a stale wrapper, the shim can inject a secret for a (pwd, command) pair that the on-disk config no longer maps.

## Evidence / mechanism

- The shim fetches values from the agent, not from disk: `internal/shim/shim.go` → `fetchEnv` → `cli.FetchEnvCwd(ctx, pwd, cmd)` (~line 491). The agent resolves against its **in-memory** config/index.
- Reload is only wired to the reload op: `ag.SetReload(loadAndBuild)` in `internal/cli/unlock.go` and `internal/cli/agent_internal.go`. Callers that ping it: `internal/tui` (after save) and `internal/cli/clone.go` (`pingAgentReloadFromClone`). There is no `reload` subcommand and no fs-watch.
- `is-mapped` and `__chpwd` read the file fresh every call, so the **routing** decision is always current; only the agent's **value resolution** is stale.

## Suggested fixes (pick one)

1. **Agent stat-on-fetch, reload-if-changed (most robust).** The agent records the config mtime at load; on each `fetch_env` / `fetch_env_cwd` (or on a cheap timer) it stats the file and re-runs `loadAndBuild` if the mtime advanced. Self-healing for any editor. Needs the master key in memory (already held while unlocked) so re-decrypt works. Watch for: races with a half-written file (mtime + size, or only reload on a stable mtime), and the AtomicSave rename semantics (new inode — stat by path, not fd).
2. **fsnotify watch on the config path.** Event-driven; slightly more moving parts (rename/replace means re-arming the watch on the parent dir).
3. **`jitenv reload` CLI command (cheapest).** A manual escape hatch that sends the existing reload op. Doesn't auto-heal, but documents the workflow and is trivial. Could ship alongside (1) as the explicit form.

## Related

- #201 — fixes the shell command-hash staleness layer; references this as a follow-up.
- The third layer (reconcile prompt-fire timing) is tracked separately.

Galvill:
## Summary

Closes #202.

A running agent only reloaded its config on an explicit reload op (TUI save / `clone`) or a `lock`/`unlock` cycle — there is no `jitenv reload` command and no file-watch. So a `config.toml` edited **by hand or by any external tool** was never seen: the agent kept serving env values from its stale in-memory config. Combined with the cwd-mapping flow, the shim could inject for a `(pwd, command)` pair the on-disk config no longer maps, or miss a freshly-added one.

This is the agent-side staleness layer found while investigating the cwd-mapping issues (the shell-command-hash layer was fixed in #201).

## Fix

The agent now records the config mtime (`SetConfigPath`) and, before serving the **data ops** (`is_mapped` / `fetch_env` / `fetch_env_cwd` / `cwd_commands`), stats the file and reloads via the existing `SetReload` callback when the mtime has advanced.

- **mtime-gated**: the common case is a single `stat` whose result matches the cached mtime → returns immediately. No re-decrypt / rebuild unless the file actually changed.
- **serialised** (`reloadMu`) with a double-check, so concurrent fetches reload at most once per change.
- **fail-safe**: a failed reload (e.g. a transient half-written file) leaves the previous resolver in place and does not advance the cached mtime, so it's retried on the next op rather than wedging the agent on a broken config.
- Wired at both daemon entry points — `__agent` (`internal/cli/agent_internal.go`) and `unlock --foreground` (`internal/cli/unlock.go`) — via the new `Agent.SetConfigPath`.

The TUI / `clone` reload ping is unchanged; it's now just a fast-path for the same outcome (the agent would pick the edit up on the next fetch anyway).

## Tests

`TestAgentAutoReloadsOnConfigMtimeChange` (`internal/agent/agent_test.go`), at the same in-process socket+client level as the existing `TestAgentReload`:

- Pre-edit fetch returns the first resolver's value; reload callback not called.
- After editing the file + bumping its mtime, the next fetch returns the new value.
- A further fetch with no new edit does **not** reload again — asserts the reload is mtime-gated (exactly one reload call).

## Test plan

- [x] `go test ./internal/agent/ ./internal/cli/` — pass.
- [x] `go build ./...`, `go vet ./internal/agent/ ./internal/cli/`, `gofmt -l` — clean.
- [x] **Negative check**: with the `maybeReload()` dispatch hook removed, `TestAgentAutoReloadsOnConfigMtimeChange` fails; restored, it passes — confirming the test exercises the fix.

## Notes

- The real-crypto reload path (`loadAndBuild`: `config.Load` → `DecryptInPlace` → `Validate` → `BuildResolver`) is the same closure already used by the explicit reload op and exercised by the daemonize/wrapper e2e at startup; this PR only changes *when* it's invoked.
- The remaining follow-up (#203, the `__chpwd` reconcile one-prompt timing window) is the lowest-severity layer and is left as-is per its issue discussion.

### DIFF
--- a/internal/agent/agent_test.go
+++ b/internal/agent/agent_test.go
@@ -167,6 +167,67 @@ func TestAgentReload(t *testing.T) {
 	}
 }
 
+// TestAgentAutoReloadsOnConfigMtimeChange covers #202: once SetConfigPath
+// is set, a data op picks up an out-of-band edit to the config (mtime
+// advanced) without an explicit reload op. The reload is mtime-gated, so
+// it fires exactly once per change — not on every fetch.
+func TestAgentAutoReloadsOnConfigMtimeChange(t *testing.T) {
+	first := &fakeResolver{env: map[string]map[string]string{"/a": {"K": "v1"}}}
+	second := &fakeResolver{env: map[string]map[string]string{"/a": {"K": "v2"}}}
+
+	cfgFile := filepath.Join(newTestAgentDir(t), "config.toml")
+	if err := os.WriteFile(cfgFile, []byte("# v1\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	var reloadCalls int
+	a, p := newTestAgent(t, first)
+	a.SetReload(func() (Resolver, error) {
+		reloadCalls++
+		return second, nil
+	})
+	a.SetConfigPath(cfgFile)
+
+	go a.Serve(context.Background()) //nolint:errcheck // goroutine: error surfaced via Shutdown/Listen pair
+	cli := NewClient(p.Socket)
+	deadline := time.Now().Add(2 * time.Second)
+	for time.Now().Before(deadline) {
+		if _, err := cli.Status(context.Background()); err == nil {
+			break
+		}
+		time.Sleep(20 * time.Millisecond)
+	}
+
+	// No edit yet → first resolver, no reload.
+	if env, _ := cli.FetchEnv(context.Background(), "/a"); env["K"] != "v1" {
+		t.Fatalf("pre-edit K=%q (want v1)", env["K"])
+	}
+	if reloadCalls != 0 {
+		t.Fatalf("pre-edit reloadCalls=%d (want 0)", reloadCalls)
+	}
+
+	// Edit the file and bump its mtime past the recorded baseline.
+	if err := os.WriteFile(cfgFile, []byte("# v2\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	future := time.Now().Add(2 * time.Second)
+	if err := os.Chtimes(cfgFile, future, future); err != nil {
+		t.Fatal(err)
+	}
+
+	// Next fetch auto-reloads → second resolver.
+	if env, _ := cli.FetchEnv(context.Background(), "/a"); env["K"] != "v2" {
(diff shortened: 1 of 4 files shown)


Galvill:
Closing without merging — not pursuing the auto-reload approach.

Design decision: `config.toml` is only expected to be edited through jitenv's own commands (`jitenv config` / the TUI, and `jitenv clone`), all of which already ping the agent's reload op after a successful save. Hand-editing the file directly (or editing it with an external tool) is not a supported workflow, so the agent serving "stale" in-memory config in that case is acceptable — the auto-reload-on-mtime machinery adds a stat to every data op plus a reload path to cover a scenario we don't support.

See #202 for the full rationale.

Issue #258, opened 2026-06-04
bash hook: per-prompt fork storm — DEBUG trap forks `jitenv is-mapped` for every PROMPT_COMMAND-internal command (regression from #237)


Galvill:
## Symptom

A noticeable delay was added to **every prompt** after #237 (PR #242) shipped. With `JITENV_HOOK_DEBUG=1`:

```
jitenv-hook: candidate cmd=[jitenv __chpwd "$$" ...] resolved=[/usr/bin/jitenv]
jitenv-hook: is-mapped rc=1
jitenv-hook: candidate cmd=[[ -z "${ZSH_VERSION-}" ]] resolved=[/usr/bin/[]
jitenv-hook: is-mapped rc=1
jitenv-hook: candidate cmd=[git rev-parse --git-dir --is-inside-git-dir ...] resolved=[/usr/bin/git]
jitenv-hook: is-mapped rc=1
jitenv-hook: candidate cmd=[[ -z "$repo_info" ]] resolved=[/usr/bin/[]
jitenv-hook: is-mapped rc=1
```

Every command the **git-prompt** runs during prompt redraw forks a `jitenv is-mapped` (process spawn + full TOML parse, now larger post-#248).

## Root cause

[...]

Issue #248, opened 2026-06-03
config: encrypt source-map / bag-map / bag-key names via opaque IDs + sealed name-map in [_meta]


Galvill:
## Today's behavior

#235 sealed every *value* under `mappings[*].vars[*]` (env var names, sources, refs, keys, extras, literals) and the per-bag secret values. But TOML *map keys* — the structural names that bracket each table — remained plaintext on disk:

```toml
[sources.vault]                       # <-- "vault" plaintext
  type = "local"

[secrets.stripe]                      # <-- "stripe" plaintext
  "SECRET_KEY" = "enc:v2:..."         # <-- "SECRET_KEY" plaintext
  "PUBLISHABLE_KEY" = "enc:v2:..."    # <-- ditto
```

This was the carved-out non-goal in the #235 issue body (verbatim): *"source-map/bag-map keys stay plaintext"*. The carve-out exists because TOML map keys are structural, not values — `EncryptInPlace`'s `map[string]any` / `map[string]string` walks (`internal/config/encrypt.go:25-100`) only touch values. Replacing the keys themselves with envelope strings is a syntax-level change with knock-on effects across the resolver, TUI, AAD construction, and `Validate`.

## What this leaks today

A passive reader of `config.toml` (backup, screenshot, dotfiles repo, OS file sync) sees:

- **Which sources you have configured** and their type. `[sources.prod_aws] type = \"aws\"` says \"this user has AWS prod credentials.\"
- **Which bags exist and what kind of secrets they hold**. `[secrets.stripe]` keyed by `\"SECRET_KEY\"` / `\"PUBLISHABLE_KEY\"` / `\"WEBHOOK_SIGNING_SECRET\"` enumerates the user's Stripe integration shape down to which keys are payment- vs webhook-related — without ever decrypting a single value.
- **Which sources each var pulls from**. Even after #235 sealed `var.source`, the `[sources.<name>]` plaintext key still anchors the topology — an attacker who reads the file can pair var-shape inference (`enc:v2:` of 32–40 bytes → likely an API key) with the named source it points to.

This is the same threat model #235 addressed for var fields; closing it here finishes the job.

## Proposed shape (sketch)

Replace user-facing names with opaque IDs in the TOML structure, store a sealed `name-map` in `[_meta]` that maps IDs back to human-readable names. The agent / TUI dereferences IDs ↔ names at load / display time; the on-disk file is name-free.

```toml
version = 1

[_meta]
  kdf = \"argon2id\"
[...]

Issue #237, opened 2026-06-03
hook: trigger path/glob mappings on PATH-resolved bare-name commands (not just absolute / ./ / ../)


Galvill:
## Today's behavior

The bash/zsh DEBUG-trap hook in `internal/shell/snippets/{bash,zsh}.sh` only routes a command through `jitenv is-mapped` when the first token starts with `/`, `./`, or `../`. Bare-name commands resolved via `$PATH` are silently skipped — see `bash.sh:167-174`:

```bash
if [[ "$first" = /* ]]; then
    resolved="$first"
elif [[ "$first" = ./* || "$first" = ../* ]]; then
    resolved="$(cd "$(dirname "$first")" 2>/dev/null && pwd)/$(basename "$first")"
else
    return 0     # <-- everything else short-circuits here
fi
```

So given a `path = "/usr/local/bin/terraform"` (or a matching `glob = "/usr/local/bin/*"`) mapping and a user who types just `terraform`, the hook never calls `is-mapped`. Env vars don't get injected. The user has to type `$(command -v terraform)` or the full path to get the mapping to fire.

## The gap

Two of the three mapping kinds (`path`, `glob`) are file-identity-based — they say *\"whenever this specific file runs, inject these vars.\"* But today they only fire on explicit-path invocations, which is the less common way most users invoke their tools. The `cwd_glob` kind exists partly to cover the bare-name case (it materializes per-shell wrappers under `$__JITENV_WRAP_DIR` for any command name listed in its `commands`), but `cwd_glob` is *directory-scoped*. There's no clean way today to say *\"intercept terraform globally, regardless of cwd, regardless of how it was typed.\"*

Concrete examples that don't work today:
- Mapping `path = \"/usr/local/bin/terraform\"`, user types `terraform`.
- Mapping `glob = \"/usr/local/bin/kubectl-*\"`, user types `kubectl-foo`.
- Mapping `path = \"/opt/homebrew/bin/aws\"` on macOS, user types `aws`.

In each case the resolved-via-PATH absolute path *would* match the mapping if `is-mapped` were given it, but the hook never asks.

## Proposed

When the first token is a bare name (no `/`, no `./`, no `../`) and the command actually resolves through `$PATH` to a real file, resolve it and call `is-mapped` against the resolved absolute path. Same dispatch as the existing two branches (rc 0 → `jitenv run`, rc 1 → run normally, rc 2 → run normally).

Sketch for the bash hook:

```bash
[...]

PR #259, opened 2026-06-04
fix(shell): stop bash hook forking is-mapped during PROMPT_COMMAND (#258)

Galvill:
## Problem

A long delay was added to **every prompt** after #237 (PR #242). The bash hook's `DEBUG` trap runs under `shopt -s extdebug` (functrace), so it fires for every command bash runs *inside* `PROMPT_COMMAND` while redrawing the prompt — the user's git-prompt (`git rev-parse …`, `[[ … ]]`) and jitenv's own `jitenv __chpwd`.\n\nPre-#237 those bare names short-circuited with no subprocess. #237 added the bare-name `type -P` branch that **forks `jitenv is-mapped`** for every PATH-resolvable command, so every prompt redraw became a fork storm — one `jitenv` spawn + full TOML parse (larger post-#248) per prompt-internal command, every prompt.\n\nObserved (`JITENV_HOOK_DEBUG=1`): `jitenv __chpwd`, `[[ … ]]`, `git rev-parse …` each emit `candidate cmd=… → is-mapped` on every prompt.\n\nzsh is **not** affected (its `zle accept-line` widget only fires on interactive Enter).\n\n## Fix\n\nLatch the DEBUG trap off for the duration of `PROMPT_COMMAND` (the bash-preexec pattern): `__jitenv_prompt_begin` (prepended) sets `__JITENV_IN_PROMPT`, `__jitenv_prompt_end` (appended) clears it, and the trap early-returns while it's set. The existing chpwd entry is bracketed too. Interactively-typed commands are still intercepted (the latch is clear when the user's line runs).\n\nBaked PROMPT_COMMAND becomes: `__jitenv_prompt_begin;__jitenv_chpwd;<existing>;__jitenv_prompt_end`.\n\n## Verification\n\n- New regression test `TestBashHookSkipsPromptCommandCommands`: a bare command run inside PROMPT_COMMAND produces **no** `is-mapped` dispatch; the same command typed interactively **does**.\n- Manual repro (before): prompt redraw forks `is-mapped` for the prompt tool. After: zero forks during redraw, interactive interception preserved.\n- `go test ./...` green; `go vet ./...` clean; full `internal/shell` + `internal/run` e2e suites pass.\n\n## Follow-up (separate)\n\nEven with the latch, an interactively-typed bare-name command forks `is-mapped` once. #237's deferred mitigation (gate the bare-name branch on whether any `path`/`glob` mapping exists, cached from `__chpwd`) would zero that out for cwd_glob-only users — worth a separate PR.\n\nCloses #258

### DIFF
--- a/internal/shell/hook_prompt_command_test.go
+++ b/internal/shell/hook_prompt_command_test.go
@@ -1,0 +1,102 @@
+//go:build !windows
+
+package shell_test
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/gv/jitenv/internal/config"
+)
+
+// TestBashHookSkipsPromptCommandCommands is a regression test for the
+// per-prompt fork storm (issue #258). The bash DEBUG trap runs under
+// `extdebug` (functrace), so it fires for every command bash executes
+// while redrawing the prompt — including a git-prompt's `git rev-parse`
+// / `[[ … ]]` and jitenv's own `jitenv __chpwd`. After issue #237 made
+// the bare-name branch fork `jitenv is-mapped`, those prompt-internal
+// commands each spawned a `jitenv` process on every prompt.
+//
+// The hook brackets PROMPT_COMMAND with a latch (__jitenv_prompt_begin /
+// __jitenv_prompt_end) so the trap skips prompt-internal commands while
+// still intercepting interactively-typed ones. This test asserts both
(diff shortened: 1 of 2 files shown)


Galvill:
Superseded by #262. The PROMPT_COMMAND latch here was fragile — it only suppresses prompt commands that sit inside jitenv's bracket, and breaks when the prompt is registered after the hook (reproduced). #262 fixes the actual root cause: the trap no longer forks `is-mapped` for commands that can't match (in-shell anchor pre-filter), which is robust regardless of PROMPT_COMMAND ordering and works for path/glob users too.

